PR #18, opened 2025-10-02
ci: add license check job

### LINKED ISSUE #17: Add dependency license checking to CI

andyjakubowski:
Fixes #17 

This PR introduces automated license compliance checks using license-checker-rseidelsohn.
- Adds a license-check script in package.json to enforce an allowlist of permissive licenses (MIT, Apache-2.0, BSD, ISC).
- Excludes the root deepnote package from scanning to avoid false positives (UNLICENSED due to "private": true).
- CI will now fail if any dependency with a disallowed license is introduced.

This ensures that all dependencies in the monorepo remain compatible with our Apache-2.0 licensing.

## Summary by CodeRabbit

* **Chores**
  * Introduced automated license compliance checks into CI to ensure dependencies use approved licenses.
  * Added a project script and development dependency to run the license audit locally and in CI.
  * Minor spelling-dictionary update for tooling names.
  * No user-facing functionality changes; improves governance and build reliability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,4 +183,4 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Run license checker
-        run: pnpm exec license-checker-rseidelsohn --json --onlyAllow "MIT;Apache-2.0;BSD-2-Clause;BSD-3-Clause;ISC" --excludePackages "deepnote"
+        run: pnpm run license-check


.github/workflows/ci.yml
```diff
@@ -161,3 +161,26 @@ jobs:
...
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run license checker
+        run: pnpm exec license-checker-rseidelsohn --json --onlyAllow "MIT;Apache-2.0;BSD-2-Clause;BSD-3-Clause;ISC" --excludePackages "deepnote"
```
jamesbhobbs: Use the package.json script here rather than dupe the config.

package.json
```diff
@@ -15,6 +15,7 @@
     "build": "pnpm -r run build",
     "format": "biome format --write .",
     "format:check": "biome format .",
+    "license-check": "license-checker-rseidelsohn --json --onlyAllow \"MIT;Apache-2.0;BSD-2-Clause;BSD-3-Clause;ISC\" --excludePackages \"deepnote\"",
```
jamesbhobbs, line 18: @coderabbitai is this the full list of licenses usable in an apache 2 licensed monorepo?

.github/workflows/ci.yml
andyjakubowski: Done

package.json
andyjakubowski: I thought of allowing more licenses, but it seems like the vast majority of deps use one of these common licenses, and others could create a compliance risk.

Here’s a generated summary:

> Allowing more licenses upfront reduces CI noise, but it increases legal/compliance risk.
> 	•	Strict list (what you have now):
> 	•	✅ CI only passes the most common, well-understood permissive licenses.
> 	•	✅ Any edge-case license triggers review, so nothing slips in unnoticed.
> 	•	⚠️ Occasional overhead when a new, but actually safe, license (e.g. 0BSD) pops up.
> 	•	Broad list (everything permissive):
> 	•	✅ Almost zero CI failures, smooth dev workflow.
> 	•	⚠️ Higher chance you allow in a license that has odd obligations or edge cases (e.g. ECL-2.0, Zlib, Artistic-2.0) without review.
> 	•	⚠️ Harder to demonstrate diligence if you ever need to audit or defend your license posture.
> 
> Idiomatic OSS best practice is to start with the conservative core (MIT, Apache, BSD, ISC) and only expand if/when you encounter real packages under other safe licenses. That way:
> 	•	You minimize ongoing management (most deps “just work”).
> 	•	You keep a safety net to catch surprises.
> 
> So “allow more right away” feels faster short-term, but in practice most teams prefer conservative now + add when needed. That way you don’t trade off legal safety for convenience.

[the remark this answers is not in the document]
andyjakubowski: Yep already fixed that

jamesbhobbs: @coderabbitai do you agree with Andy?

PR #25, opened 2025-10-03
feat: Add support for creating Python and Markdown content from Deepnote blocks.

Artmann:
## Summary by CodeRabbit

- **New Features**
  - Added a new package to generate Markdown and Python code from notebook blocks, covering text, separators, images, code, SQL, visualizations, big numbers, buttons, and full input types (text, textarea, checkbox, select, slider, file, date, date range).
  - Utilities for safe Python string/variable handling and SQL environment name resolution; date-range helpers included.
- **Documentation**
  - Added package README with overview, supported block types, and usage examples.
- **Tests**
  - Added comprehensive unit tests for markdown, Python generation, and input block handling.
- **Chores**
  - Expanded spell-check dictionary with relevant terms.

### DIFF
--- a/packages/blocks/package.json
+++ b/packages/blocks/package.json
@@ -4,8 +4,17 @@
   "description": "",
   "keywords": [],
   "license": "Apache-2.0",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "type": "module",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs",
+      "types": "./dist/index.d.ts"
+    }
+  },
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
   "files": [
     "dist",
     "README.md",
@@ -13,8 +22,9 @@
     "package.json"
   ],
   "scripts": {
-    "build": "tsdown",
-    "test": "vitest"
+    "build": "tsdown --format esm --format cjs --dts",
+    "test": "vitest",
+    "watch": "tsdown --watch --format esm --format cjs --dts"
   },
   "dependencies": {
     "ts-dedent": "^2.2.0"


packages/blocks/src/python-code.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (isButtonBlock(block)) {
+    return createPythonCodeForButtonBlock(block, executionContext)
+  }
+
+  if (isBigNumberBlock(block)) {
+    return createPythonCodeForBigNumberBlock(block)
+  }
+
+  throw new UnsupportedBlockTypeError(`Creating python code from block type ${block.type} is not supported yet.`)
+}
```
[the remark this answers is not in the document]
Artmann: We don't have to overcomplicate this.

packages/blocks/src/markdown.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      type: 'text-cell-p',
+      content: 'Regular paragraph',
+      metadata: {},
+      sortingKey: 'a0',
+    }
+
+    const result = stripMarkdown(block)
+
+    expect(result).toEqual('Regular paragraph')
+  })
```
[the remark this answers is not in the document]
Artmann: I've fixed this

Artmann:
@coderabbitai can you review again?

jankuca:
It's currently dead code right? I'm not seeing any usage of the `createPythonCode` function.

Artmann:
> It's currently dead code right? I'm not seeing any usage of the `createPythonCode` function.

Yeah, it will be used in the extensions.

jankuca:
Ok, what are the next steps when it comes to unifying this with the deepnote-internal logic?

Artmann:
> Ok, what are the next steps when it comes to unifying this with the deepnote-internal logic?

Once we are happy with this, we should be able to use it from our internal app as well. The way we map the blocks against the database is slightly different, so I haven't included it here; we'll have to handle that separately. 

So, deepnote.com and the Code extension will have their own code to handle the blocks and persist them, but we can use this package to generate the Python and Markdown code that the Deepnote Kernel will use.

jankuca:
Great, so the goal is to replace our internal block->python logic with this package. We should do that ASAP to avoid any potential drift between these implementations. The code already looks a little different I notice.

packages/blocks/src/blocks.ts
```diff
@@ -0,0 +1,28 @@
+export interface DeepnoteBlock {
+  blockGroup: string
+  executionCount?: number
+  id: string
+  sortingKey: string
+  type: string
+}
+
+export type BlockMetadata = Record<string, unknown>
```
jankuca: Is this deliberately loosely typed without a strict discriminated union with keys based on the supported block types?

jankuca: I do see that this is being extended by the specific subtypes but we are still using this type in functions like createPythonCode(). I would expect that such functions would get a union of the specific types, not the permissive base type.

Artmann: Yeah, that's a good point. I'll change that :)

Artmann: Updated

andyjakubowski:
I opened a stacked PR to add config for tsdown build outputs, so that the `@deepnote/blocks` can be installed in other projects like `vscode-deepnote` or `jupyterlab-deepnote` for local development: https://github.com/deepnote/deepnote/pull/27

PR #30, opened 2025-10-07
feat: Add configuration for publishing @deepnote/blocks to GitHub.

Artmann:
## Summary by CodeRabbit

- Chores
  - Optimized package distribution by excluding development and test artifacts from published packages for smaller downloads and cleaner installs.
  - Updated package metadata, configured publishing to the GitHub Packages registry, and bumped the package version to 1.1.0 to streamline releases.
  - No changes to user-facing functionality or public APIs.

### DIFF
--- a/packages/blocks/package.json
+++ b/packages/blocks/package.json
@@ -5,7 +5,8 @@
   "keywords": [],
   "repository": {
     "type": "git",
-    "url": "https://github.com/deepnote/deepnote.git"
+    "url": "https://github.com/deepnote/deepnote.git",
+    "directory": "packages/blocks"
   },
   "license": "Apache-2.0",
   "type": "module",


packages/blocks/package.json
```diff
@@ -30,5 +34,8 @@
     "ts-dedent": "^2.2.0",
     "yaml": "^2.8.1",
     "zod": "^4.1.12"
+  },
+  "publishConfig": {
```
jamesbhobbs, line 39: Does this work automatically or does there need to be a ci script for publishing?

Artmann: I'm publishing it manually for now, but yes we should add a CI script for it.

Artmann:
> My understanding is this PR does the following:
> 
> * Adds `.npmignore` to exclude patterns from getting published to a registry.
> * Adds `repository` and `publishConfig` fields which are used as metadata in registries, and are picked up by publishing software.
> 
> It does not:
> 
> * Do actual publishing to GitHub — this would happen in CI, as you pointed out in another comment?
> 
> If that’s true, then perhaps we could edit this PR’s title to reflect the scope and merge it. And then follow up with a CI publishing PR?

That's correct! I've updated the title, and I'll look at the publishing workflow :)

PR #34, opened 2025-10-08
feat: Add table state to Code and SQL blocks.

Artmann:
## Summary by CodeRabbit

- New Features
  - Code and SQL blocks now persist and apply table view settings (sorting, filters, pagination, column order, hidden columns, display names, conditional formatting, wrapped text) before execution.
- Bug Fixes
  - Improved escaping of special characters in generated code and table-state payloads to prevent formatting/execution issues.
- Documentation
  - Added a comprehensive developer guide describing workflows, commands, and contribution standards.
- Tests
  - Expanded tests covering table-state serialization, escaping, and emitted code behavior.


packages/blocks/src/blocks/code-blocks.ts
```diff
@@ -10,7 +13,13 @@ export interface CodeBlock extends DeepnoteBlock {
...
 export function createPythonCodeForCodeBlock(block: CodeBlock): string {
-  return block.content
+  const dataFrameConfig = createDataFrameConfig(block)
+
+  return dedent`
+    ${dataFrameConfig}
+
+    ${block.content}
+  `
 }
```
saltenasl, line 26: Nit - i can't do a code suggestion but along the lines of:
```
export const createPythonCodeForCodeBlock = (block: CodeBlock): string => dedent`
  ${createDataFrameConfig(block)}
  ${block.content}
`
```

packages/blocks/src/blocks/data-frame.test.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    expect(result).toEqual(dedent`
+      if '_dntk' in globals():
+        _dntk.dataframe_utils.configure_dataframe_formatter('{}')
+      else:
+        _deepnote_current_table_attrs = '{}'
+    `)
+  })
+
+  it('escapes special characters in JSON string', () => {
```
saltenasl, line 144: This should be tested in the utility function instead, otherwise, if we end up removing the data-frame.ts file we will remove the tests for the utility.

saltenasl: OK, I've thought about this further and think we need the test in both places:
1. The utility function should be tested inclusive of all edge cases
2. This function needs to be tested that it is indeed using the utility to escape (because removal of it should break tests as it will break user code).

packages/blocks/src/python-code.test.ts
```diff
@@ -166,7 +166,119 @@ describe('createPythonCode', () => {
...
+            wrappedTextColumnIds: [],
+          },
+        },
+      }
+
+      const result = createPythonCode(block)
+
+      expect(result).toContain('"hiddenColumnIds":["secret"]')
     })
   })
```
saltenasl, line 311: lets please add a test that the query code is escaped

packages/blocks/src/blocks.ts
```diff
@@ -36,6 +36,8 @@ export type BlockMetadata =
   | MarkdownCellMetadata
 
 export interface ExecutableBlockMetadata {
+  [key: string]: unknown
```
jankuca: Is the purpose of this adding the data table related fields? If so, can we list them explicitly? Allowing any key here may be unnecessarily permissive.

packages/blocks/src/blocks/data-frame.ts
```diff
@@ -0,0 +1,18 @@
+import { dedent } from 'ts-dedent'
+
+import type { CodeBlock } from './code-blocks'
+import { escapePythonString } from './python-utils'
+import type { SqlBlock } from './sql-blocks'
+
+export function createDataFrameConfig(block: CodeBlock | SqlBlock): string {
+  const rawTableState = 'deepnote_table_state' in block.metadata ? block.metadata.deepnote_table_state : {}
```
jankuca: Why is the key presence check needed? Wouldn't it be enough to do `block.metadata.deepnote_table_state ?? {}`?

packages/blocks/src/blocks/data-frame.ts
```diff
@@ -0,0 +1,18 @@
...
+  const tableState = rawTableState ?? {}
+  const tableStateAsJson = JSON.stringify(tableState) ?? '{}'
```
jankuca: Why is this fallback needed? Shouldn't this always be at least `{}` before the stringifying?

(may seem like a nitpick but any fallback like this makes the reader wonder whether there is some random edge case that this is covering)

packages/blocks/src/blocks.ts
Artmann: Good point!

Artmann: I've added the real types.

packages/blocks/src/blocks/data-frame.ts
Artmann: It's not needed now that we have the explicit types. Updated!

packages/blocks/src/blocks/data-frame.test.ts
Artmann: I added the test to the `python-code` tests as well

packages/blocks/src/blocks/data-frame.ts
Artmann: Yeah that's right!

packages/blocks/src/blocks/data-frame.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      filters: [],
+      pageSize: 25,
+      pageIndex: 0,
+      columnOrder: ['date', 'value', 'category'],
+      hiddenColumnIds: [],
+      columnDisplayNames: [],
+      conditionalFilters: [],
+      cellFormattingRules: [],
+      wrappedTextColumnIds: [],
+    })
```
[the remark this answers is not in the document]
Artmann: Don't worry about it.

packages/blocks/src/blocks/data-frame.test.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    expect(result).toEqual(dedent`
+      if '_dntk' in globals():
+        _dntk.dataframe_utils.configure_dataframe_formatter('${expectedJson}')
+      else:
+        _deepnote_current_table_attrs = '${expectedJson}'
+    `)
+  })
```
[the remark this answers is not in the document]
Artmann: Don't worry about it.

packages/blocks/src/blocks/data-frame.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      filters: [{ id: 'category', value: 'A' }],
+      pageSize: 50,
+      pageIndex: 2,
+      columnOrder: ['date', 'value', 'category'],
+      hiddenColumnIds: ['_deepnote_index_column'],
+      columnDisplayNames: [{ columnName: 'value', displayName: 'Value ($)' }],
+      conditionalFilters: [],
+      cellFormattingRules: [{ column: 'value', rule: 'color' }],
+      wrappedTextColumnIds: ['category'],
+    })
```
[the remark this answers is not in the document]
Artmann: Don't worry about it.

packages/blocks/src/blocks/table-state.ts
```diff
@@ -0,0 +1,12 @@
...
+  columnDisplayNames?: { columnName: string; displayName: string }[]
+  columnOrder?: string[]
+  conditionalFilters?: unknown[]
+  filters?: { id: string; value: string }[]
+  hiddenColumnIds?: string[]
+  pageIndex?: number
+  pageSize?: number
+  sortBy?: { id: string; desc: boolean }[]
+  wrappedTextColumnIds?: string[]
+}
```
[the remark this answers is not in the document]
Artmann: Don't worry about it.

packages/blocks/src/python-code.test.ts
```diff
@@ -476,6 +616,11 @@ describe('createPythonCode', () => {
       const result = createPythonCode(block)
 
       expect(result).toEqual(dedent`
+        if '_dntk' in globals():
+          _dntk.dataframe_utils.configure_dataframe_formatter('{}')
+        else:
+          _deepnote_current_table_attrs = '{}'
+
```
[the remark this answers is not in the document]
Artmann: Don't worry about it.

PR #45, opened 2025-10-16
feat(charts): Do not attach selection for chart blocks

OlegWock:
Contributes to GRN-4778

This selection is used for interactive filters in main Deepnote app. Since we don't expose UI to filter charts in our extension at the moment, it might look confusing that user can select some part of chart but do nothing with it.

## Summary by CodeRabbit

* **Chores**
  * Added a common macOS metadata file to the project ignore list to keep repos cleaner.

* **Bug Fixes**
  * Simplified generated Python code for visualizations by removing an extraneous argument and passing the chart specification more directly, improving snippet clarity and reliability.


OlegWock:
Oh, I messed up something with .gitignore, fixing

PR #146, opened 2025-11-13
fix(database-integrations): support legacy redshift metadata with undefined auth method

saltenasl:
## Summary by CodeRabbit

* **Bug Fixes**
  * Added support for legacy Redshift integrations that lack an explicit authentication method; these now default to username and password authentication, ensuring backward compatibility with older configurations.

* **Tests**
  * Expanded test coverage for Redshift metadata validation with legacy configuration scenarios.


Artmann (approved):
🚀

PR #158, opened 2025-11-18
fix(database-integrations): update Trino auth method names to include trino- prefix, fix backward compatibility

tkislan:


tkislan:
@coderabbitai full review

PR #171, opened 2025-11-30
fix(vitest.config): enhance workspace package resolution and update alias setup

axaysagathiya:
### Problem
Running `pnpm test` was failing with the following error:
```
Error: Failed to resolve entry for package "@deepnote/blocks". 
The package may have incorrect main/module/exports specified in its package.json.
```

This occurred because workspace packages (@deepnote/blocks, @deepnote/convert, @deepnote/database-integrations) reference their built `dist/` files in `package.json` exports, but these files don't exist until the packages are built. This forced a build step before running tests.

### Solution
used `vite-tsconfig-paths` plugin to automatically resolve TypeScript path aliases defined in `tsconfig.json` file

## Summary by CodeRabbit

* **Chores**
  * Improved test environment to resolve TypeScript path aliases and workspace package imports during test runs.
  * Added tooling to enable path-alias handling in the test environment.

* **Tests**
  * Updated TypeScript project settings to set a base URL and package path mappings so tests discover workspace packages reliably.

### DIFF
--- a/packages/database-integrations/package.json
+++ b/packages/database-integrations/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepnote/database-integrations",
-  "version": "1.4.1",
+  "version": "1.4.2",
   "description": "Deepnote database integration definitions",
   "keywords": [],
   "repository": {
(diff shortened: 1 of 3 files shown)


saltenasl:
@axaysagathiya can you please do this via tsconfig aliases and `vite-tsconfig-paths` plugin as it's more future-proof

PR #184, opened 2025-12-18
feat(blocks, convert): add snapshot fields to Deepnote file format

Artmann:
Add new optional fields to support snapshot capabilities:

Top-level fields:
- `environment`: hash, pythonVersion, pythonEnvironment, platform, packages, customImage
- `execution`: startedAt, finishedAt, triggeredBy, inputs, summary, error

Block-level fields:
- `contentHash`: MD5 hash of block source code
- `executionStartedAt`: ISO 8601 timestamp when block execution started
- `executionFinishedAt`: ISO 8601 timestamp when block execution completed

Also:
- Mark `project.settings.environment` as deprecated in favor of top-level `environment`
- Preserve new fields during Deepnote <-> Jupyter roundtrip conversion
- Add tests for schema validation and roundtrip preservation

All fields are optional to maintain backward compatibility.

## Summary by CodeRabbit

* **New Features**
  * Top-level environment and execution metadata added to files and propagated into notebook/convert outputs; file metadata includes exportedAt and modifiedAt.
  * Blocks now support contentHash, executionStartedAt, executionFinishedAt; execution summaries and errors are captured. settings.environment (nested) marked deprecated in favor of top-level.

* **Tests**
  * Added coverage for environment/execution snapshots, block snapshot fields, execution error preservation, roundtrip conversion, and backward compatibility.

* **Examples**
  * Example notebooks updated to include environment and snapshot metadata.

### DIFF
--- a/examples/4_snapshot.deepnote
+++ b/examples/4_snapshot.deepnote
@@ -22,6 +22,7 @@ execution:
     totalDurationMs: 26682
 metadata:
   createdAt: '2025-12-11T10:30:00.000Z'
+  exportedAt: '2025-12-11T10:32:20.000Z'
   modifiedAt: '2025-12-11T10:32:15.123Z'
 project:
   id: snapshot-example-001


packages/blocks/src/deserialize-file/deepnote-file-schema.ts
```diff
@@ -14,7 +20,56 @@ export const deepnoteBlockSchema = z.object({
...
 export type DeepnoteBlock = z.infer<typeof deepnoteBlockSchema>
 
+export const environmentSchema = z
+  .object({
+    customImage: z.string().optional(),
+    hash: z.string().optional(),
+    packages: z.record(z.string()).optional(),
+    platform: z.string().optional(),
+    pythonEnvironment: z.enum(['uv', 'conda', 'venv', 'poetry', 'system']).optional(),
+    pythonVersion: z.string().optional(),
```
saltenasl: i've been just thinking  about `executionCount`, `executionFinishedAt` and `executionStarted` - in that case it seems like it makes sense to nest all of them under `execution` so we don't need to repeat ourselves. However, we can't easily do that because it's a breaking changes (moving `executionCount` to `execution.count`).

But then I was like OK, we should spot these early on and make the right call right out of the bat. 

So, should this be(pseudocode)
```suggestion
    python:
      environment: z.enum(['uv', 'conda', 'venv', 'poetry', 'system']).optional(),
      version: z.string().optional(),
```

so if we add more fields here it will be more structured?

Artmann: That's a good call! That also makes it easy to extend to other languages.

dinohamzic: > However, we can't easily do that because it's a breaking changes (moving executionCount to execution.count).

Just an idea, but probably overkill: could we handle this by adding a `formatVersion` property to the Deepnote format file? 

`v1` being the default (if explicitly set or absent), and `v2` the version that introduces snapshots? Then we could build custom logic depending on the version while slowing deprecating older versions.

Artmann: We do have a version field, but we would like to avoid making breaking changes if we can.

dinohamzic (approved):
Looking great. 👌

jamesbhobbs (dismissed):
1. As we're adding these fields and putting them in the validator we should also update our examples.
2. Where is the issue this is defined in? We should write it publicly to be sure there's a clear RFC?
3. I've created https://github.com/deepnote/deepnote/pull/186 which would need to be reconciled with this, it includes something useful for this PR, which is a test which runs the validation against all the .deepnote files in the repo - so the examples cannot get out of sync with the improved file format.

Artmann:
> 1. As we're adding these fields and putting them in the validator we should also update our examples.
> 2. Where is the issue this is defined in? We should write it publicly to be sure there's a clear RFC?
> 3. I've created #186 which would need to be reconciled with this, it includes something useful for this PR, which is a test which runs the validation against all the .deepnote files in the repo - so the examples cannot get out of sync with the improved file format.

@jamesbhobbs  I've updated the example files, as well as added a new example snapshot file. I've also created an issue that outlines all the work here: https://github.com/deepnote/deepnote/issues/187

dinohamzic (approved):

Issue #185, opened 2025-12-18
Define & enforce .deepnote YAML canonical constraints (parser + validator)


jamesbhobbs:
To make .deepnote files reliably portable across languages/tools and diff-friendly in git, we should standardise a strict YAML subset and enforce it in a format checker (and later fmt). Proposed hard constraints:

1. YAML 1.2 only — lock spec version to avoid YAML 1.1 implicit typing and parser discrepancies.
2. UTF-8 only — ensure consistent encoding across platforms; no BOM/odd encodings.
3. No duplicate keys — prevent ambiguous merges/overrides; duplicates must be a hard error.
4. No anchors/aliases/merge keys — avoid surprising indirection and tool-specific behaviour; keep files explicit for LLM/editor friendliness.
5. No custom tags, no implicit typing surprises — disallow !! tags and enforce safe loading; scalars shouldn’t change meaning based on parser heuristics.
6. Scalars explicitly typed by schema rules — types are determined by our schema, not YAML inference (e.g., timestamps must be quoted ISO strings; no bare dates).

Acceptance criteria:
 
- A validator that rejects violations with clear, actionable error messages (line/column where possible).
- Test corpus covering each rule (valid + invalid examples).

PR #186, opened 2025-12-18
feat: enhance deepnote file format checking strictness

jamesbhobbs:
Fixes: #185

## Summary by CodeRabbit

* **New Features**
  * Exposed a public UTF‑8 validation/decoding utility and tightened YAML parsing to reject BOMs, invalid UTF‑8, and several disallowed YAML constructs.

* **Tests**
  * Greatly expanded test coverage for string- and byte-level UTF‑8 cases, duplicate keys, anchors/aliases, custom tags, explicit typing and merge key rules.
  * Added an end-to-end suite validating workspace YAML files parse into expected structures.

### DIFF
--- a/packages/blocks/src/deserialize-file/validate-all-deepnote-files.test.ts
+++ b/packages/blocks/src/deserialize-file/validate-all-deepnote-files.test.ts
@@ -1,6 +1,6 @@
 import { readdir, readFile } from 'node:fs/promises'
 import { join, relative } from 'node:path'
-import { describe, expect, it } from 'vitest'
+import { beforeAll, describe, expect, it } from 'vitest'
 import { deserializeDeepnoteFile } from './deserialize-deepnote-file'
 
 /**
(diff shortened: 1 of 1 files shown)


packages/blocks/src/deserialize-file/validate-all-deepnote-files.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  // Use paths relative to workspace root
+  const workspaceRoot = join(__dirname, '../../../..')
+  const deepnoteFiles = [
+    'packages/convert/test-fixtures/All-Deepnote-blocks.deepnote',
+    'packages/convert/test-fixtures/ChartExamples.deepnote',
+    'examples/1_hello_world.deepnote',
+    'examples/2_blocks.deepnote',
+    'examples/3_integrations.deepnote',
+    'examples/demos/housing_price_prediction.deepnote',
+  ]
```
dinohamzic: What do you think about searching for `.deepnote` files recursively from the root folder, instead of having to keep this list manually up to date?

jamesbhobbs: @dinohamzic done

packages/blocks/src/deserialize-file/parse-yaml.ts
```diff
@@ -1,8 +1,131 @@
...
+ *
+ * @param bytes - Raw file bytes as Uint8Array
+ * @returns Decoded UTF-8 string without BOM
+ * @throws Error if BOM detected or invalid UTF-8 encoding
+ *
+ * @example
+ * ```typescript
+ * const bytes = await fs.readFile('file.deepnote') // Returns Buffer/Uint8Array
+ * const yamlContent = decodeUtf8NoBom(bytes)
+ * const parsed = parseYaml(yamlContent)
```
dinohamzic, line 15: Would it be better DX to expose a single `parseYamlFromBytes` function that encapsulated these 2 steps?

jamesbhobbs: Claude's answer, I don't mind...

Looking at the usage patterns, I'd say it's not worth it for these reasons:
Current State Analysis
Current usage:
decodeUtf8NoBom() is publicly exported from the package index
parseYaml() is used internally by deserializeDeepnoteFile()
All actual usage in the codebase is deserializeDeepnoteFile(yamlContent) which takes a string
Key observations:
The two-step pattern (decodeUtf8NoBom → parseYaml) is only documented, not actually used anywhere in the codebase
deserializeDeepnoteFile() already provides the high-level API users actually need
decodeUtf8NoBom() was exported to give users control over byte-level validation when needed
Why not combine them:
Separation of concerns: UTF-8 decoding and YAML parsing are fundamentally different operations
Flexibility: Some users might want to decode UTF-8 bytes but not parse YAML (e.g., just validate encoding)
String sources: Many use cases already have strings (from APIs, databases, etc.) and don't need byte-level validation
Testing: Easier to test each function independently
Current API is already simple: deserializeDeepnoteFile(string) is the main entry point
Better DX improvement:
If anything, I'd suggest removing decodeUtf8NoBom from the public exports since:
It's not used anywhere in the actual codebase
The documentation example is theoretical
Users who have bytes can just do new TextDecoder('utf-8', { fatal: true }).decode(bytes) themselves
The main API deserializeDeepnoteFile() already handles everything
Verdict: Keep the current design. The separation is clean, and deserializeDeepnoteFile() already provides the simple high-level API users need.

PR #194, opened 2025-12-23
chore: prepush runs lint format typecheck test spell-check

jamesbhobbs:
## Summary by CodeRabbit

* **Chores**
  * Enhanced pre-push validation with automated code quality, type checking, testing, and spelling verification to maintain code standards during development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,12 +76,13 @@ Reusable packages and libraries powering Deepnote's notebook, runtime, and colla
 - **Markdown conversion**: Convert text blocks to/from markdown format
 - **Input block support**: Text, textarea, checkbox, select, slider, file, date, and date-range inputs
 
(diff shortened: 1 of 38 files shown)


.husky/pre-push
```diff
@@ -3,3 +3,5 @@ if [ "$branch_name" = "main" ]; then
   echo "Pushing directly to main is not allowed."
   exit 1
 fi
+
+pnpm lintAndFormat && pnpm typecheck && pnpm test && pnpm spell-check
```
Artmann, line 7: Does it make a difference if we run these in parallel?

Artmann (approved):

.husky/pre-push
jamesbhobbs: The output becomes gibberish.

jamesbhobbs: They also return 0 before the others complete.

jamesbhobbs: The right option would be to wrap them in something like turborepo without caching and let it handle the exit code overall.

PR #215, opened 2026-01-14
feat(@deepnote/convert): add per-block Jupyter export with convertBlockToJupyterCell

saltenasl:
## Summary

- Add `convertBlockToJupyterCell` function for streaming export of individual Deepnote blocks to Jupyter cells
- Export the function and `JupyterCell` type for TypeScript consumers

## Motivation

Consumers need to export large notebooks to Jupyter format in a streaming fashion—processing blocks one at a time rather than loading the entire notebook into memory.

## Usage

```typescript
import { convertBlockToJupyterCell } from '@deepnote/convert'
import type { JupyterCell } from '@deepnote/convert'

for await (const block of blockStream) {
  const cell: JupyterCell = convertBlockToJupyterCell(block)
  outputStream.write(JSON.stringify(cell))
}
```

## Release Notes

* **New Features**
  * New capability to convert individual Deepnote items to Jupyter format with full metadata preservation including identifiers, type information, and source data
  * Supports code, markdown, and SQL content types

* **Tests**
  * Comprehensive test coverage added for single-item and multi-item conversion scenarios

### DIFF
--- a/packages/convert/package.json
+++ b/packages/convert/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepnote/convert",
-  "version": "2.1.3",
+  "version": "2.2.0",
   "description": "",
   "keywords": [],
   "repository": {


packages/convert/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@deepnote/convert",
-  "version": "2.1.3",
+  "version": "2.2.0",
   "description": "",
```
Artmann, line 4: Was this intended?

saltenasl: yeah, i just include bumps in these to save time on waiting for approvals, release immediately after it gets merged. 

Worst case scenario - there will be a conflict in this line and i'll need to update it when it happens

PR #225, opened 2026-01-21
feat(@deepnote/convert): Create snapshots when converting notebooks.

Artmann:
- Added the `--single-file` flag to disable snapshot mode (default: `false`, meaning snapshots are enabled
  by default)
- Converting TO Deepnote now:
  - Creates a source file without outputs
  - Creates snapshots/ directory with snapshot file
- Converting FROM Deepnote now:
  - Automatically loads and merges the latest snapshot
  - Outputs are included in converted files

## Summary by CodeRabbit

* **New Features**
  * CLI flag to choose single-file (embed outputs) vs split snapshot files
  * Snapshot system: create/discover/load/split/merge with deterministic hashing and optional in-memory snapshots; write helper that can emit source + snapshot
  * Marimo support: per-cell outputs extraction and propagation into conversions
  * Read-and-convert helpers for Jupyter/Marimo/Percent/Quarto and unified, snapshot-aware conversion flow
  * Improved format detection, richer Jupyter output types, and executable-block detection

* **Tests**
  * Extensive unit and integration tests covering hashing, lookup, split/merge, format detection, Marimo outputs, and end-to-end snapshot workflows

* **Chores**
  * Added a local development CLI script


packages/convert/src/snapshot/split.ts
```diff
@@ -0,0 +1,146 @@
...
+  'input-slider',
+  'input-date',
+  'input-date-range',
+  'input-file',
+])
+
+/**
+ * Checks if a block type can have outputs
+ */
+function isExecutableBlock(type: string): boolean {
```
tkislan: This is now a duplicite function
Same logic was merged into `packages/runtime-core/src/execution-engine.ts`

packages/convert/src/snapshot/split.ts
```diff
@@ -0,0 +1,146 @@
...
+ * Creates a slug from a project name.
+ * Converts to lowercase, replaces spaces and special chars with hyphens,
+ * removes consecutive hyphens, and trims leading/trailing hyphens.
+ *
+ * @param name - The project name to slugify
+ * @returns A URL-safe slug
+ */
+export function slugifyProjectName(name: string): string {
+  return name
+    .toLowerCase()
```
tkislan, line 19: we should also do `.normalize()` or with some explicit option
as people can easily have accented characters in the project name

Artmann: Didn't know about `normalize`. Added.

packages/convert/src/snapshot/split.ts
Artmann: Nice. I've extracted it to blocks, and both places use that version now.

tkislan (dismissed):

PR #238, opened 2026-01-22
feat(@deepnote/cli): add TOON output format for LLM-optimized output

jamesbhobbs:
Add --toon flag to inspect and run commands as an alternative to --json. TOON (Token-Oriented Object Notation) reduces token usage by 30-60% compared to JSON while remaining human-readable, making it ideal for LLM consumption.

- Add @toon-format/toon dependency
- Add outputToon() function to output.ts
- Support --toon flag in inspect and run commands
- Add tests for outputToon function
- Update help text with TOON examples

## Summary by CodeRabbit

* **New Features**
  * Unified output flag (-o, --output <format>) supporting json and a new "toon" format; optional efficiency hint for toon.

* **Behavior**
  * Machine-readable output (-o json / -o toon) suppresses interactive logs and per-block rendering; validate accepts json only; invalid formats error.

* **Documentation**
  * Added "Data Formats" entries for toon/TOON/Toon and updated CLI docs/examples to use -o/--output.

* **Tests**
  * Added tests for toon output, efficiency analysis, hints, completions, and format validation.

* **Chores**
  * CLI now references an external TOON encoder dependency.

### DIFF
--- a/packages/cli/src/output.ts
+++ b/packages/cli/src/output.ts
@@ -166,7 +166,7 @@ export function outputToon(data: unknown, options?: { showEfficiencyHint?: boole
       const hint =
         savingsPercent <= 0
           ? `Hint: JSON would be ${Math.abs(savingsPercent).toFixed(0)}% smaller for this data. ` +
-            'Consider --json for better compatibility.'
+            'Consider --output json for better compatibility.'
           : `Hint: TOON only saves ${savingsPercent.toFixed(0)}% over JSON for this data. ` +
             'TOON works best with uniform arrays of objects (30-60% typical savings).'
 


packages/cli/src/commands/inspect.ts
```diff
@@ -24,6 +25,8 @@ export function createInspectAction(
       const exitCode = error instanceof FileResolutionError ? ExitCode.InvalidUsage : ExitCode.Error
       if (options.json) {
```
tkislan: Or what if we would change the option to `output` enum with currently available enum options of "json" and "toon"
e.g. like `kubectl` has it https://kubernetes.io/docs/reference/kubectl/#formatting-output
or also `aws`, `gcloud`, `docker` and other cli tools as well

tkislan (commented):
I would consider merging the output flags into a single `-o/--output` one

packages/cli/src/commands/inspect.ts
dinohamzic: That's a great point, it makes way more sense to do it as you are proposing. Will apply the changes in a sec.

dinohamzic: Changes applied here: https://github.com/deepnote/deepnote/pull/238/changes/3800aeb1cf024584f2d140468d7cc22d4f5a51f6

packages/cli/src/cli.ts
```diff
@@ -197,7 +224,13 @@ ${c.bold('Exit Codes:')}
...
     .description('Validate a .deepnote file against the schema')
     .argument('<path>', 'Path to a .deepnote file to validate')
-    .option('--json', 'Output in JSON format for scripting')
+    .option('-o, --output <format>', 'Output format: json', (value: string): OutputFormat => {
+      // Validate command only supports JSON output (no TOON)
+      if (value !== 'json') {
+        throw new Error(`Invalid output format: ${value}. Valid formats: json`)
+      }
+      return value as OutputFormat
+    })
```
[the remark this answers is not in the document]
dinohamzic: Where is this code duplicated?

packages/cli/src/output.ts
tkislan: this is now outdated

packages/cli/src/cli.ts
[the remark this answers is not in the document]
dinohamzic: Fixed.

packages/cli/src/output.ts
dinohamzic: Nice catch. Fixed!

PR #254, opened 2026-01-28
feat(cli): add analyze command and LLM context features

jamesbhobbs:
- Add shared analysis utilities (analysis.ts) extracting reusable functions from lint.ts and stats.ts for stats, lint, and DAG analysis
- Add new `analyze` command for comprehensive project analysis including quality score, structure info, dependency analysis, and suggestions
- Add `--context` flag to run command that includes analysis info (stats, lint issues, variable definitions/usages) in output
- Add auto-diagnosis on failure: when JSON/TOON output is enabled and execution fails, automatically include upstream dependencies, related lint issues, and used variables to help LLMs understand failures
- Refactor lint.ts and stats.ts to use shared analysis utilities, reducing code duplication by ~550 lines

## Summary by CodeRabbit

* **New Features**
  * New analyze command to inspect .deepnote files (human/JSON/TOON outputs), with shell completions and notebook/python options.
  * Run command gains --context to include per-block and project analysis and diagnostics.
  * Automatic execution snapshotting saves run outputs for later inspection.
  * Input widget types and SQL environment helpers exposed for integrations.

* **Tests**
  * Extensive tests added for analysis, diagnostics, stats, completions, and snapshot persistence.

### DIFF
--- a/packages/convert/package.json
+++ b/packages/convert/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@deepnote/convert",
-  "version": "2.3.0",
+  "version": "2.3.1",
   "description": "",
   "keywords": [],
   "repository": {


test-fixtures/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Ignore snapshot directories created by tests
```
jamesbhobbs, line 1: @dinohamzic we should probably undo this, commit the snaphots and also use them for tests in ci. But this is out of scope for this PR

packages/cli/src/commands/analyze.ts
```diff
@@ -0,0 +1,379 @@
...
+
+async function analyzeFile(path: string | undefined, options: AnalyzeOptions): Promise<AnalyzeResult> {
+  const { absolutePath } = await resolvePathToDeepnoteFile(path)
+
+  debug('Reading file contents...')
+  const rawBytes = await fs.readFile(absolutePath)
+  const yamlContent = decodeUtf8NoBom(rawBytes)
+
+  debug('Parsing .deepnote file...')
+  const deepnoteFile = deserializeDeepnoteFile(yamlContent)
```
dinohamzic, line 74: no action: I'm just making a note to myself to refactor things like this (also eg. error handling, --output flag shared logic, etc.) to single function invocations, as we are duplicating similar logic in multiple commands

I'll dedicate a bit of time to this after this fast PRs cycle slows down a bit

packages/cli/src/utils/analysis.ts
```diff
@@ -0,0 +1,858 @@
...
+
+// ============================================================================
+// Constants
+// ============================================================================
+
+/** Built-in integrations that don't require external configuration */
+const BUILTIN_INTEGRATIONS = new Set(['deepnote-dataframe-sql', 'pandas-dataframe'])
+
+/** Input block types */
+const INPUT_BLOCK_TYPES = new Set([
```
dinohamzic: This should ideally come from @deepnote/blocks

packages/cli/src/commands/analyze.ts
jamesbhobbs: "I'll dedicate a bit of time to this after this fast PRs cycle slows down a bit" I hope it won't slow down :)

packages/cli/src/utils/analysis.ts
```diff
@@ -0,0 +1,858 @@
...
+    blockCount: notebook.blocks.length,
+    linesOfCode: totalLoc,
+    blockTypes,
+  }
+}
+
+/**
+ * Count lines of code in a block's content.
+ */
+function countLinesOfCode(block: DeepnoteBlock): number {
```
dinohamzic, line 662: We should be able to easily extend ast-analyzer.py (in the reactivity package) to return LOC as well, instead of having to manually compute them like this.

packages/cli/src/utils/analysis.ts
```diff
@@ -0,0 +1,858 @@
...
+    return loc
+  }
+
+  return content.split('\n').filter(line => line.trim()).length
+}
+
+/**
+ * Extract imported module names from a code block.
+ */
+function extractImports(block: DeepnoteBlock): string[] {
```
dinohamzic: Also this, we should see if we can get this from the reactivity lib.

packages/cli/src/utils/output-persistence.test.ts
```diff
@@ -0,0 +1,273 @@
...
+                content: '# Header',
+                sortingKey: 'a2',
+                blockGroup: 'group-1',
+                metadata: {},
+              },
+            ],
+          },
+        ],
+      },
+    }) as DeepnoteFile
```
dinohamzic: Should we use fixtures instead?

packages/cli/src/utils/output-persistence.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @param sourcePath - Path to the source file
+ * @param file - The DeepnoteFile
+ * @returns The snapshot file path
+ */
+export function getSnapshotPath(sourcePath: string, file: DeepnoteFile): string {
+  const snapshotDir = getSnapshotDir(sourcePath)
+  const slug = slugifyProjectName(file.project.name) || 'project'
+  const snapshotFilename = generateSnapshotFilename(slug, file.project.id, 'latest')
+  return resolve(snapshotDir, snapshotFilename)
+}
```
dinohamzic, line 133: @Artmann is snapshot related logic in this file something that we can find in another package? 

if not, I suppose it would be good to move this outside of `packages/cli`, as we will need to reuse it in the extension? (but can be done later)

packages/cli/src/cli.ts
```diff
@@ -543,6 +545,37 @@ ${c.bold('Examples:')}
     })
     .action(createStatsAction(program))
 
+  // Analyze command - comprehensive project analysis
+  program
```
dinohamzic, line 549: We need to update completions for `analyze`.

packages/cli/src/utils/output-persistence.ts
Artmann: Yeah, we have code for writing and reading snapshots in the extension, and we have code for reading them in convert. It would be good to have it together in one place :)

Artmann: https://github.com/deepnote/vscode-deepnote/blob/main/src/notebooks/deepnote/snapshots/snapshotService.ts

Artmann: https://github.com/deepnote/deepnote/tree/main/packages/convert/src/snapshot

dinohamzic: Thanks Chris. I'm giving this a try but it's a big change. I'll merge this PR as is, and then I'll open a new one with the refactor immediately after.

PR #263, opened 2026-01-30
feat: Add `llm` as an output option

Artmann:
Add `--output llm` as a first-class output format option across all commands. The llm format is an abstraction that resolves to the best format for LLM consumption:

- Commands supporting toon: llm → toon
- Commands with JSON-only: llm → json

This allows the underlying implementation to change in the future without users needing to update their commands.

## Summary by CodeRabbit

* **New Features**
  * Added "llm" as a supported output format across many commands (inspect, run, cat, convert, open, validate, dag, stats, lint, diff, analyze, completion, etc.), preserving existing defaults.
  * Updated command help texts and shell completions (Bash, Zsh, Fish) to list and validate the new format.

* **Tests**
  * Added/updated tests to validate "llm" format resolution and error messaging.

### DIFF
--- a/packages/cli/src/commands/run.test.ts
+++ b/packages/cli/src/commands/run.test.ts
@@ -2,7 +2,8 @@ import fs from 'node:fs'
 import os from 'node:os'
 import { join } from 'node:path'
 import { Command } from 'commander'
-import { afterEach, beforeEach, describe, expect, it, type Mock, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, type Mock, type MockedFunction, vi } from 'vitest'
+import type { saveExecutionSnapshot } from '../utils/output-persistence'
 
 // Create mock engine functions
 const mockStart = vi.fn()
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli.ts
```diff
@@ -135,7 +135,11 @@ function registerCommands(program: Command): void {
     .command('inspect')
     .description('Inspect and display metadata from a .deepnote file')
     .argument('[path]', 'Path to a .deepnote file or directory (defaults to current directory)')
-    .option('-o, --output <format>', 'Output format: json, toon', createFormatValidator(OUTPUT_FORMATS))
+    .option(
+      '-o, --output <format>',
+      'Output format: json, toon, llm',
```
dinohamzic, line 140: I feel that from a purely user perspective this is a bit confusing: there is nothing indicating that `llm` is just an alias for either `toon` or `json`, leaving me wondering which one should I use.

Not sure what the solution could be, maybe writing it as `Output format: json, toon, llm (alias for toon)`? (and similarly in the other places)

Artmann: The idea is that we don't want to expose that to the user. If you are using an LLM, you should use the `llm` option, not have to worry about what it is.

PR #272, opened 2026-02-09
fix: run dependent upstream blocks when you run a single block

jamesbhobbs:
NB consider whether to make this optional

## Summary by CodeRabbit

* **New Features**
  * Running blocks can include resolved upstream dependencies; CLI dry-run shows upstream ordering.
  * Support running multiple target block IDs together; explicit list takes precedence over single-block target.
  * Notebook-scoped upstream analysis (when notebook not specified, scope is confined).

* **Bug Fixes / UX**
  * Clearer errors when blocks are missing or non-executable; graceful handling of partial/missing upstream deps.

* **Tests**
  * Added tests for upstream resolution, multi-block execution, precedence, filtering, and edge cases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,6 +195,33 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: ./coverage/lcov.info
           fail_ci_if_error: true
+  npm-publish-dry-run-mcp:
+    name: NPM Publish Dry Run (@deepnote/mcp)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4
+
+      - name: Setup Node.js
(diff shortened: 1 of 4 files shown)


packages/reactivity/src/dag-analyzer.ts
```diff
@@ -39,3 +39,43 @@ function getDownstreamBlocks(dag: BlockDependencyDag, blocksIds: string[], visit
...
+    nodes: dag.nodes.filter(node => node.order <= maxOrder),
+  }
+
+  return getUpstreamBlocks(filteredDag, blocksIds)
+}
+
+function getUpstreamBlocks(dag: BlockDependencyDag, blocksIds: string[], visited: Set<string> = new Set()): string[] {
+  const inputVariables = dag.nodes.filter(node => blocksIds.includes(node.id)).flatMap(node => node.inputVariables)
+
+  const blocksThatDefineVariables = dag.nodes
```
Artmann, line 66: How does this work? Will this only find blocks where a variable is defined? Don't we need to walk up the three and grab all the blocks along the way?

For examle if we have three blocks:

```
A
df = pd.Dataframe()

B
df["a"] = 42

C 
print(df)
```

If we run `C`, we would like to execute `A -> B -> C` to get the correct result.

jamesbhobbs: They are all found by DAG and C has dependencies A+B. Execution order is currently based on order in the notebook and thus it will run A->B->C.

PR #285, opened 2026-02-12
perf: skip upstream blocks with cached outputs when running a single …

jamesbhobbs:
…block

When running with --block, upstream dependencies with valid cached outputs (matching contentHash in the latest snapshot) are now skipped instead of being re-executed every time. Adds --force-upstream-blocks flag to bypass the cache and force re-execution of all upstream blocks.

Performance improvement of https://github.com/deepnote/deepnote/pull/272 

## Release Notes

* **New Features**
  * Introduced `--force-upstream-blocks` CLI option for the run command, enabling forced re-execution of all upstream blocks and bypassing the output cache
  * Upstream blocks now intelligently skip re-execution when cached outputs match the current content, unless the force flag is used


jamesbhobbs:
> As discussed via DM, this approach breaks running blocks with upstream dependencies as we still don't support stateful runs.
> 
> > The cached upstream blocks are being skipped entirely, they don't execute at all. But those blocks define Python variables (like import numpy as np) that downstream blocks depend on. The snapshot only stores cell outputs (printed text, images), not the kernel state (variables in memory). So when we skip an upstream block, its variables never exist in the kernel, and the downstream block fails with NameError.

The proper solution would be a persistent kernel daemon, which is a separate, larger effort.

PR #303, opened 2026-02-19
fix(cli): apply --input overrides to input block metadata

jamesbhobbs:
## Summary

- Fix bug where `--input` CLI values were ignored because input blocks unconditionally overwrote them with saved metadata during execution
- Add `applyInputOverrides()` that patches the in-memory `DeepnoteFile` before execution so input blocks use CLI-provided values
- Add test documenting that empty projects (no executable blocks) return exit code 0

## Test plan

- [x] `applyInputOverrides` patches matching input blocks
- [x] Non-matching input blocks are left untouched
- [x] Empty inputs object is a no-op
- [x] Works across multiple notebooks
- [x] Skips non-input blocks (e.g., code blocks)
- [x] Empty project returns exit code 0
- [x] All 1911 existing tests pass
- [x] `pnpm typecheck` passes
- [x] `pnpm biome:check` passes

## Summary by CodeRabbit

* **New Features**
  * CLI input parameters are applied to interactive input blocks prior to execution, allowing command-line overrides to set input values.

* **Tests**
  * Added comprehensive tests verifying overrides update matching input blocks, leave non-matching/non-input blocks untouched, and are a no-op when empty.

* **Examples**
  * Added an example notebook demonstrating multiple input types (text, slider, checkbox) and their overridden values.

### DIFF
--- a/examples/6_with_inputs.deepnote
+++ b/examples/6_with_inputs.deepnote
@@ -1,0 +1,62 @@
+metadata:
+  createdAt: '2026-02-20T00:00:00.000Z'
+project:
+  id: input-overrides-example
+  name: Input overrides example
+  notebooks:
+    - blocks:
+        - blockGroup: g1
+          content: ''
+          id: input-greeting
+          metadata:
+            deepnote_variable_name: greeting
+            deepnote_input_label: Greeting
+            deepnote_variable_value: Hello from the file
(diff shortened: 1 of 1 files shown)


dinohamzic:
@jamesbhobbs changes work nicely. I just committed the .deepnote file I used for testing as an example file.

PR #314, opened 2026-02-20
fix(cli): resolve skill directory relative to executable for pypi installs

saltenasl:
Bun-compiled binaries freeze import.meta.url at build time, causing install-skills to look for skills at the CI build path. Resolve relative to process.execPath as a fallback, and ship skills alongside the binary in the pypi wheel.

## Summary by CodeRabbit

* **Refactor**
  * Skills are now properly bundled and included in PyPI package distributions
  * Improved skill directory resolution for better compatibility across bundled packages, system-level installations, and development environments


cspell.json
```diff
@@ -55,13 +55,15 @@
     "deepnote",
     "dntk",
     "dockerfiles",
-    "ename",
-    "execv",
-    "evalue",
-    "extglob",
     "dockerised",
     "dropna",
```
saltenasl, line 59: I just used the IDE command to add the new word, not sure which part of the pipeline sorted these

PR #315, opened 2026-02-20
feat(@deepnote/cli): show better imports in analyze command

dinohamzic:
Screenshot 2026-02-23 at 10 07 23

## Summary by CodeRabbit

* **New Features**
  * Import analysis now extracts and tracks actual package names (e.g., pandas, numpy) rather than module aliases, improving visibility into project dependencies across CLI output and analysis reports.

* **Chores**
  * Bumped reactivity package version to 1.2.0.

### DIFF
--- a/packages/cli/src/utils/analysis.ts
+++ b/packages/cli/src/utils/analysis.ts
@@ -73,7 +73,6 @@ export interface ProjectStats {
   blockTypesSummary: BlockTypeStats[]
   notebooks: NotebookStats[]
   imports: string[]
-  packageAliases: Record<string, string>
 }
 
 export interface BlockInfo {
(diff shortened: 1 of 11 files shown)


dinohamzic:
@coderabbitai approve

packages/cli/src/utils/analysis.ts
```diff
@@ -730,20 +732,38 @@ function countLinesOfCode(block: DeepnoteBlock): number {
...
     }
   }
   return Array.from(imports).sort()
 }
 
+/**
+ * Extract package-to-alias mapping from DAG nodes.
+ * Only includes explicit "as" renames (e.g. pandas → pd from "import pandas as pd").
+ */
+function extractPackageAliasesFromDag(dag: BlockDependencyDag): Record<string, string> {
```
OlegWock: This incorrectly parses `import os.path as paaaaath` as `Imports: os (as paaaaath)`

OlegWock (changes requested):
Code looks good, but there is one issue with incorrect parsing of aliases 👀

packages/cli/src/utils/analysis.ts
dinohamzic: I simplified the PR and removed the alias altogether (going back to your original suggestion), otherwise it's becoming a bit of a headache. 😄

PR #425, opened 2026-07-20
fix(database-integrations): throw typed error for malformed integrations YAML

### LINKED ISSUE #424: mergeApiIntegrationsIntoYaml throws "Document with errors cannot be stringified" on a malformed existing integrations file
## Summary

`mergeApiIntegrationsIntoYaml(existingContent, apiIntegrations)` in `@deepnote/database-integrations` throws `Error: Document with errors cannot be stringified` when `existingContent` is a **readable but malformed** YAML file — e.g. a file containing git merge-conflict markers, or a YAML typo. Instead of regenerating the file, the call crashes.

Because this is the primitive that merges API integrations into the user's **local** integrations file (its own doc comment says it "mirrors what `deepnote integrations pull` does to the YAML file"), a `deepnote integrations pull` run against a locally-corrupted file would fail hard rather than heal it.

## Reproduction

Against `@deepnote/database-integrations@1.5.0` (node 22, `yaml@2.8.3`):

```js
const { mergeApiIntegrationsIntoYaml, parseIntegrationsDocument } = require('@deepnote/database-integrations')

const conflict = [
  'integrations:',
  '<<<<<<< HEAD',
  '  - id: a',
  '=======',
  '  - id: b',
  '>>>>>>> branch',
  '',
].join('\n')

mergeApiIntegrationsIntoYaml(conflict, [])              // throws
mergeApiIntegrationsIntoYaml('a: b: c\n', [])           // throws (plain YAML typo)
mergeApiIntegrationsIntoYaml('integrations: []\n', [])  // OK
parseIntegrationsDocument(conflict)                     // returns a Document with errors.length === 6 (NOT null)
```

Observed output:

```
mergeApiIntegrationsIntoYaml(conflict-markers, []) => THREW: Error: Document with errors cannot be stringified
mergeApiIntegrationsIntoYaml("a: b: c", [])        => THREW: Error: Document with errors cannot be stringified
mergeApiIntegrationsIntoYaml("integrations: []", []) => OK
parseIntegrationsDocument(conflict)                => Document, errors.length=6
```

## Root cause

`parseIntegrationsDocument` uses `yaml.parseDocument(...)`, which **collects** parse errors onto `doc.errors` rather than throwing or returning `null`; it only returns `null` for empty/whitespace content:

```js
function parseIntegrationsDocument(content) {
  if (!content.trim()) return null
  return parseDocument(content, { strict: true, version: '1.2' })
}
```

`mergeApiIntegrationsIntoYaml` then does:

```js
const doc = (existingContent != null ? parseIntegrationsDocument(existingContent) : null) ?? createNewDocument()
// ... mergeApiIntegrationsIntoDocument(doc, apiIntegrations) ...
return { content: serializeIntegrationsDocument(doc), /* ... */ }
```

For malformed input, `parseIntegrationsDocument` returns a **non-null** errored `Document`, so `?? createNewDocument()` never fires. `serializeIntegrationsDocument` is `doc.toString()`, and `yaml`'s `Document.toString()` throws `Document with errors cannot be stringified` when `doc.errors.length > 0`.

(A related case: a well-formed file whose `integrations:` value is not a sequence throws `InvalidIntegrationsTypeError` via `getOrCreateIntegrationsFromDocument`.)

## Impact

- Any consumer that merges into a possibly-corrupt existing file crashes instead of self-healing. The most realistic trigger is a **git merge conflict** in a version-controlled integrations file (or a manual edit typo) — precisely the situation for a file that lives in a user's repo.

## Suggested fix

Treat a malformed/errored existing document as "start fresh" instead of trying to serialize it. In `mergeApiIntegrationsIntoYaml`, fall back to `createNewDocument()` when the parsed document is `null` **or** has `doc.errors.length > 0` (and handle the non-sequence `integrations:` case). Alternatively, have `parseIntegrationsDocument` return `null` on parse errors so the existing `?? createNewDocument()` fallback already covers it.

## Notes / caveats

- The CLI/extension user-facing impact is **inferred** from the doc comment on `mergeApiIntegrationsIntoYaml` ("mirrors what `deepnote integrations pull` does to the YAML file"). I verified the library primitive throws; I did not trace the exact `pull` call path, so it's worth confirming whether `pull` already wraps this.

<sub>Filed with an automated reproduction via Claude Code.</sub>

tkislan:
Fixes #424

## Problem

`mergeApiIntegrationsIntoYaml(existingContent, apiIntegrations)` failed with the cryptic `Error: Document with errors cannot be stringified` when `existingContent` was readable-but-malformed YAML — most commonly a file left with unresolved git merge conflict markers.

`yaml.parseDocument()` never throws; it returns a `Document` with `.errors` populated. So the `?? createNewDocument()` fallback never fired, the errored document flowed through the whole merge, and the failure only surfaced at serialize time — with no indication of which file or which line was broken.

The same defect reached the CLI through `readIntegrationsDocument`, affecting `integrations pull`, `add` and `edit`. `pull` and `add` hit the serialize-time crash; `edit` failed earlier with a *misleading* semantic error (`Integration with ID "…" not found`), because its `isSeq`-guarded helpers silently returned empty on the errored document.

## Approach

Fail fast at parse time behind a typed, exported error. No silent self-healing — the user is shown the path and the parse detail, and recovery is manual.

- **`parseIntegrationsDocument`** throws `IntegrationsYamlParseError` when `doc.errors` is non-empty. Contract is now: `null` for blank, `Document` for valid, throws for malformed.
- **The CLI** wraps it as `MalformedIntegrationsFileError`, carrying the file path plus the parse detail (line, column, code frame), and maps it to **exit code 2** (invalid usage) in all three commands.
- **`pull`** now reads the local file *before* the empty-response early return, so a malformed file is caught on every run — not just when the workspace has integrations to pull.

Both error classes are exported so downstream consumers (webapp git-sync, the VS Code extension) can catch them and rebuild via `mergeApiIntegrationsIntoYaml(null, apiIntegrations)` if they'd rather discard a corrupt file than surface the error.

### Before / after

```
$ deepnote integrations edit some-id --file broken.yaml

# before
Error: Document with errors cannot be stringified          (exit 1)

# after                                                     (exit 2)
Invalid YAML in integrations file: broken.yaml

Implicit keys need to be on a single line at line 2, column 1:

integrations:
<<<<<<< HEAD
^

This usually comes from unresolved merge conflict markers (<<<<<<<, =======, >>>>>>>) or a manual-edit typo.
Open the file, fix the reported line, and re-run the command — or delete the file manually if you no longer need its contents.
```

## Why `doc.errors` is the right gate

This moves a failure earlier on a shared code path, so it's worth being precise about the blast radius.

`Document.toString()` refuses on `if (this.errors.length > 0)` (`yaml/dist/doc/Document.js:322`) — **the same predicate** as the new gate. The set of documents that now throw at parse time is therefore identical *by construction* to the set that already threw at serialize time. Nothing that works today starts failing; a cryptic late failure becomes an actionable early one.

Verified empirically across the boundary cases: duplicate keys, tab-as-indent and bad scalar starts all already threw at serialize time (so they improve), while CRLF, trailing tabs and `postgres://h:5432/db` produce zero errors and are untouched. Warnings are deliberately *not* gated on — `TAG_RESOLVE_FAILED` (e.g. `a: !custom v`) lands in `doc.warnings` and stringifies fine; gating on those would have been a real regression.

**One intentional behavior change:** `pull` against a zero-integration workspace with a malformed local file previously exited 0 without ever reading the file. It now exits 2. This is deliberate, tested, and documented.

**Bonus fix:** `updateDotEnv` previously ran *before* the failing serialize, so a malformed file could leave a half-written `.env` behind. The relocated read closes that window — the new tests assert `.env` is never created.

## Tests

26 new tests. `mergeApiIntegrationsIntoYaml` had **zero** coverage before this, so it gets baseline coverage alongside the regression cases.

- **`integrations-document.test.ts`** (new) — blank → `null`, valid → round-trips, conflict markers and `a: b: c` → typed error exposing `.errors` and line info.
- **`merge-into-yaml.test.ts`** (new) — fresh-document and comment-preserving merge snapshots, both malformed fixtures, `InvalidIntegrationsTypeError`, and a test that *executes* the documented catch-and-rebuild recovery recipe. One test asserts the message does **not** contain `Document with errors cannot be stringified`, pinning the actual regression.
- **CLI** — `readIntegrationsDocument` unit coverage, plus end-to-end cases for all three commands asserting exit code 2, the YAML file byte-identical afterwards, and `.env` never created. The `edit` action test pins **both** branches of the new ternary (malformed → 2, missing file → 1).

## Known gap (pre-existing, not addressed here)

An unresolved YAML alias (`password: *missing`) produces `errors: 0, warnings: 0` yet still throws at `toString()`. Same shape as this issue — late crash, cryptic message, partial `.env` write — but a different trigger, and behavior is identical before and after this change. Worth a follow-up issue; out of scope here.

## Verification

| Gate | Result |
| --- | --- |
| `pnpm test` | 2524 passed / 144 files |
| `pnpm run typecheck` | clean |
| `pnpm lintAndFormat` | clean |
| `pnpm spell-check` | clean |
| `pnpm build` | clean |
| CLI smoke | malformed → exit 2 + file unmodified; missing → exit 1 |

Docs updated per `AGENTS.md`: new `skills/deepnote/references/cli-integrations.md` (registered in `SKILL.md`) and a note under `integrations pull` in the CLI README.

https://claude.ai/code/session_01PyawjygDS8h5qdX1ZWJoRH

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed local integrations YAML, including unresolved merge conflicts, with clearer recovery guidance.
  * `integrations pull`, `add`, and `edit` now exit with code **2** for invalid YAML without modifying integrations or secret files.

* **Documentation**
  * Updated integrations command documentation with new defaults, shared options, interactive behavior, secret handling, and invalid-YAML guarantees.
  * Added an integrations CLI reference and updated the quick-reference table.

* **Tests**
  * Expanded coverage for YAML validation, error details, exit codes, and file preservation.


tkislan:
@coderabbitai full review

m1so (dismissed):

jamesbhobbs (approved):
Approving diff since last review not the whole PR

PR #448, opened 2026-08-05
fix(cloud): harden cloud runs and add --storage-mode

jamesbhobbs:
## Summary

Combines cloud-run reliability fixes with the `--storage-mode` feature (formerly #433).

**Reliability fixes:**
- **Shared HTTP layer** — new `packages/cloud/src/http.ts` with schema-validated requests, replacing ad-hoc fetch calls
- **Request deadlines** — mandatory timeout that can't be disabled by a caller abort signal
- **Bounded snapshot polling** — `waitForRunSnapshot` replaces duplicated polling logic across CLI and local-runner; empty snapshot content is treated as not-yet-attached, never written out as an empty file
- **Artifact vs execution separation** — execution status and artifact delivery are reported independently (`status` vs `artifactStatus`/`artifactError`)
- **No-op snapshot synthesis** — local empty/markdown-only notebooks get a valid output-free snapshot, explicitly marked `artifactStatus: "synthesized"` and disclosed in human output

**`--storage-mode` feature (from #433):**
- Adds `--storage-mode read-write|readonly` to `deepnote run --cloud`
- Maps to the API's `detachedRunStorageMode` field (as recommended in Dino's review)
- Shell completions for bash, zsh, and fish
- Guard: `--storage-mode` cannot combine with `--block` (block runs use live mode)

### `-o json` result shape (not a behavior break)

`success` keeps the meaning it had on `main`: the run succeeded **and** its snapshot was delivered. Exit codes are likewise unchanged: `0` = full success, `1` = any failure (run or artifact). What changed is where artifact problems are reported: snapshot retrieval/save failures now land in the dedicated `artifactStatus`/`artifactError` fields instead of the run-level `error` field, which now carries execution errors only. `artifactStatus` is one of `saved`, `synthesized` (no API artifact; an output-free snapshot was written from the local source), `not_produced`, or `unavailable`.

## Test plan

- [x] All 1,248 cloud + CLI tests pass locally
- [x] Full repo test suite (2,916 tests) passes
- [x] CI passes after rebase
- [ ] Human review

## Summary by CodeRabbit

- **New Features**
  - Added `--storage-mode` support for detached cloud runs, with `read-write` and `readonly` options.
  - Added reliable polling for delayed snapshot delivery, including cancellation and timeout handling.
  - Added synthesized snapshots for empty or markdown-only notebooks.
  - Improved reporting to distinguish execution success, artifact production, and artifact availability.
  - Added clearer machine-readable errors and shell completions for storage-mode values.

- **Documentation**
  - Updated cloud and CLI guidance for storage modes, snapshot behavior, and artifact outcomes.


packages/cloud/src/cloud-runs.ts
```diff
@@ -477,17 +453,109 @@ export async function fetchSnapshotContent(run: NormalizedRun, options: FetchSna
...
+
+    if (attempt >= attempts) {
+      if (lastReadError !== undefined) {
+        throw lastReadError
+      }
+      return { run: current, content: null }
+    }
+
+    if (attempt > 0) {
+      await sleep(intervalMs)
```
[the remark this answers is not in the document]
jamesbhobbs: Fixed in 212f6ed. Caller-provided sleeps are now raced against the abort signal with listener cleanup, and the regression test aborts while a custom sleep remains permanently pending.

packages/cloud/README.md
```diff
@@ -41,23 +41,32 @@ is in preview and its exact shape may drift. Failures throw `ApiError`
...
+| `describeRunError(run)`                                                                   | A human-readable message for a failed run, if the API supplied one.                                              |
+| `isTerminalStatus` / `isSuccessStatus` / `isFailedStatus`                                 | Status classifiers. Unknown statuses are treated as non-terminal, so a drifting API cannot hang.                 |
+| `RUN_STATUSES`, `RunStatus`                                                               | The known run statuses.                                                                                          |
+| `RunTimeoutError`                                                                         | Thrown when `pollRunUntilComplete` exceeds its deadline. Carries the `runId` — the run may still be executing.   |
+| `NormalizedRun`, `TriggerRunBody`, `GetRunOptions`, `PollOptions`, `FetchSnapshotOptions` | Types.                                                                                                           |
```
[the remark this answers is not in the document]
jamesbhobbs: Fixed in 212f6ed. The README now documents the SettledRunSnapshot return contract, explains that content may be null, and lists SettledRunSnapshot and WaitForRunSnapshotOptions among the exported types.

jamesbhobbs:
Addressed the outside-diff snapshot-test finding in 212f6ed as well. The no-op CLI test now reads the saved snapshot, asserts its exact serialized content, verifies fixed execution timestamps and preserved source blocks, and proves stale outputs/execution counts are removed. Full validation: 2,790 tests passed (1 skipped), typecheck, Biome, Prettier, and CSpell all passed.

packages/cli/src/utils/run-in-cloud.test.ts
```diff
@@ -405,23 +422,61 @@ describe('runInDeepnoteCloud — output and exit codes', () => {
...
+    await runInDeepnoteCloud(path, {
+      cloud: true,
+      token: 't',
+      url: API_URL,
+      output: 'json',
+      notebook: 'No-op',
+    })
 
     const logged = (console.log as unknown as ReturnType<typeof vi.fn>).mock.calls.at(-1)?.[0] as string
     const result = JSON.parse(logged)
```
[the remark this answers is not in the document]
jamesbhobbs: Narrowed both `JSON.parse(logged)` calls to `CloudRunResult` via explicit type annotation, and imported the type. All 29 tests pass. Please resolve this thread if satisfied.

packages/cli/src/cli.test.ts
```diff
@@ -168,6 +169,11 @@ describe('CLI', () => {
       expect(output).toContain('COMPREPLY')
       expect(output).toContain('schedule')
       expect(output).toContain('--timezone')
+      expect(output).toContain(`    # Handle --storage-mode option completion for detached cloud runs
+    if [[ "\${prev}" == "--storage-mode" && "\${subcommand}" == "run" ]]; then
+        COMPREPLY=( $(compgen -W "read-write readonly" -- "\${cur}") )
+        return 0
+    fi`)
```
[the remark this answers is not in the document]
jamesbhobbs: Added exact assertions for `--storage-mode` completion output in both the Zsh and Fish tests — matching the generated strings from `completions.ts`. All 23 CLI tests pass. Please resolve if satisfied.

packages/cli/src/utils/run-in-cloud.ts
```diff
@@ -375,31 +416,31 @@ export async function runInDeepnoteCloud(path: string | undefined, options: RunC
...
-    if (success) {
-      snapshotError = `Failed to save snapshot: ${message}`
-    } else {
-      debug(`Failed to save snapshot for failed run ${finalRun.runId}: ${message}`)
-    }
+    artifactStatus = 'unavailable'
+    artifactError = `Failed to retrieve or save snapshot: ${message}`
+  }
+
+  if (success && options.out && artifactStatus === 'not_produced') {
```
voyti: One main Agent finding: exit signal here might be misleading, creating a false success. A notebook with real code, whose snapshot lags past ~3s now reports success with no artifact unless --out was given. It contradicts the old explicit contract, has no test pinning it as intended, and bites automation silently.

packages/cli/src/utils/run-in-cloud.ts
```diff
@@ -355,16 +385,27 @@ export async function runInDeepnoteCloud(path: string | undefined, options: RunC
...
+  // and a valid no-op run may never produce an artifact at all.
   let snapshotPath: string | undefined
   let timestampedSnapshotPath: string | undefined
-  let snapshotError: string | undefined
+  let artifactStatus: CloudArtifactStatus = 'not_produced'
+  let artifactError: string | undefined
   try {
-    const content = await fetchSnapshotContent(finalRun, { baseUrl, token })
-    if (content) {
+    const settled = await waitForRunSnapshot(baseUrl, token, finalRun, {
```
voyti, line 400: Agent-found, optional but might be beneficial - plumb WaitForRunSnapshotOptions (at least sleep/intervalMs) through runInDeepnoteCloud  — it removes ~12s of real sleeping from the test suite and lets you skip/shorten the ~3s settle for known no-ops and failed runs.

voyti (commented):
General comment about the release notes possible valuable addition (agent found): flag the `-o json` success semantics change as breaking for machine consumers (exit codes are the compatible signal; artifactStatus is the new source of truth).

packages/cli/src/utils/run-in-cloud.ts
jamesbhobbs: This is already guarded — lines 378–393 force `commandSucceeded = false` (and `exitCode = 1`) whenever the run succeeds but the snapshot is missing or fails to persist. The test at line 408 ("fails a successful run that returns no snapshot content") pins this exact scenario: `success: false`, `status: "success"`, `error` matching `/no snapshot content/`, no `snapshotPath`, exit 1.

The `renderHumanResult` path at line 429 is only reached when `result.success` is true, which requires both a successful run *and* a persisted snapshot — so `snapshotPath` is always set here. The conditional is defensive, not a real branch.

Is there a specific scenario you see that isn't covered by the existing guard? Happy to add a test if so. Please resolve if satisfied.

packages/cli/src/utils/run-in-cloud.ts
jamesbhobbs: The `installFetch` mock already returns terminal status on the first GET, so `pollRunUntilComplete` never enters a sleep cycle — the test suite has no real sleeping from polling. The injectable `sleep`/`now` on `PollOptions` are already available in `@deepnote/cloud` and used in the cloud-runs tests, so if a CLI-level test ever needs them they can be plumbed through at that point. Keeping the surface minimal for now since no caller needs it. Happy to revisit if you see a concrete test that would benefit. Please resolve if satisfied.

jamesbhobbs:
@voyti Good call — I've updated the PR description with a "⚠️ Breaking: `-o json` success semantics" section that documents the change: `success` now means run succeeded AND snapshot persisted. Exit codes (`0`/`1`) are unchanged and remain the backward-compatible signal.

packages/cli/src/utils/run-in-cloud.ts
voyti: Okay, please see is this clarification helps (looks like your verification was done on origin/main?): 
"the lines you're citing (378–393, the test at 408, the renderHumanResult branch at 429) all match origin/main — this PR removes that guard and that test, and at HEAD success is run-status only (:436), so run-in-cloud.test.ts:579 now asserts success: true, artifactStatus: 'not_produced' with exit 0 for a successful run with no snapshot. That's the false-success case for any code notebook whose snapshot lags past the ~3s settle window without --out, and the "Breaking" note in the description currently describes the opposite of this behavior."

packages/cli/src/utils/run-in-cloud.ts
voyti: (sent by agent due to Github UI outage)

The ~12 s isn't from polling — it's `waitForRunSnapshot`'s default 1.5 s ×2 settle sleep, which the CLI can't inject; that's why the four snapshot-less tests needed `10_000` timeouts and each takes ~3.0 s (they drop to ~10 ms with the interval stubbed). Not blocking, but plumbing `sleep`/`intervalMs` through `runInDeepnoteCloud` (or a settle option) removes it and keeps this file from being the suite's long pole.

packages/cloud/src/cloud-runs.ts
```diff
@@ -477,17 +464,134 @@ export async function fetchSnapshotContent(run: NormalizedRun, options: FetchSna
...
+        snapshotDelivery: 'inline',
+        requestTimeoutMs: options.requestTimeoutMs,
+        signal: options.signal,
+      })
+    } catch (error) {
+      if (options.signal?.aborted) {
+        throw options.signal.reason
+      }
+      options.onRetryError?.(error)
+    }
```
voyti, line 600: (sent by agent due to Github UI outage)

If every settle re-fetch fails (401/404/5xx), the errors are swallowed via `onRetryError` and the function still returns `{ content: null }`, which callers read as "the run produced no snapshot". An API outage should not look like a valid empty run — please surface the last re-fetch error instead, and have the CLI skip snapshot synthesis in that case.

packages/cli/src/utils/run-in-cloud.ts
jamesbhobbs: Fixed. My previous reply cited `origin/main` code — apologies. At HEAD, `result.success` was raw run status (line 436), so a successful run with no snapshot reported `success: true` and exit 0. Now `result.success` uses `commandSucceeded`, which is `false` when `artifactError` is set. The `options.out` guard is removed — `artifactError` is set whenever a successful run produces no snapshot, regardless of `--out`. Updated the test from `ExitCode.Success` to `ExitCode.Error` for this case. Please resolve if satisfied.

packages/cli/src/utils/run-in-cloud.ts
jamesbhobbs: Correcting my previous reply — the ~12s is from `waitForRunSnapshot`'s settle loop (1.5s × attempts), not polling. The CLI doesn't currently pass `sleep`/`intervalMs` through to `waitForRunSnapshot`. Since you marked this non-blocking, I'll address the settle-sleep plumbing in a follow-up to keep this PR focused. Please resolve if satisfied.

packages/cloud/src/cloud-runs.ts
jamesbhobbs: Fixed. The `onRetryError` callback now tracks `lastRetryError`. After `waitForRunSnapshot` returns and no-op synthesis is attempted, if content is still null and `lastRetryError` is set, it's thrown — the catch block sets `artifactStatus = 'unavailable'` with a clear error message instead of silently leaving it as `not_produced`. A known no-op run with re-fetch failures still synthesizes its snapshot correctly (existing test covers this). Added a regression test for the all-failed-re-fetches scenario. Please resolve if satisfied.

skills/deepnote/references/cli-run.md
```diff
@@ -88,19 +89,33 @@ The notebook to run is resolved in this order:
...
-saved also exits `1` (with `success: false` and an `error`).
+`runId`, `status`, and any `snapshotPath`. `success` describes notebook execution;
+`artifactStatus` separately reports `saved`, `not_produced`, or `unavailable`.
+
+After terminal status, the CLI polls briefly for snapshot attachment. A successful empty or
+markdown-only notebook can legitimately produce none: with a local file, the CLI synthesizes a
+valid output-free snapshot from that source; with only `--notebook-id`, it exits `0` with
+`artifactStatus: not_produced`. An advertised snapshot that cannot be downloaded or saved reports
+`artifactStatus: unavailable`, includes `artifactError`, and exits `1`. `--out` also exits `1` when
+no artifact was produced because the explicitly requested path cannot be written.
```
voyti: (agent observation) The docs say success only means "the notebook ran fine" and that a run with no snapshot exits 0. The code does neither: success is false whenever the snapshot isn't saved, and no-snapshot exits 1. Please update the docs to match the code. Also, the PR description's "breaking change to success" isn't right — main already worked this way; the real change is that snapshot errors moved from error to artifactError.

packages/cli/src/utils/run-in-cloud.ts
```diff
@@ -199,6 +205,33 @@ function parseSnapshotContent(content: string): DeepnoteSnapshot | null {
...
+  const executed: DeepnoteFile = {
+    ...source,
+    execution: {
+      startedAt: run.createdAt ?? run.completedAt ?? fallbackTime,
+      finishedAt: run.completedAt ?? fallbackTime,
+      triggeredBy: 'api',
+    },
+  }
+  return serializeDeepnoteSnapshot(splitDeepnoteFile(executed).snapshot)
+}
```
voyti, line 237: (agent observation)  When the API returns no snapshot for a markdown-only local notebook, the CLI now writes a snapshot it made up from the local file and reports it exactly like a real one (saved, exit 0). If the local file is out of date, or the API is just down (the test at :536 passes with every re-fetch returning 503), the user gets a wrong "no outputs" file and no warning. Please mark these differently (e.g. artifactStatus: 'synthesized') and say so in the terminal output.

A bit similar thing in cloud-runs.ts:562, run-in-cloud.ts:411 -  if the API returns snapshot content that is an empty string, the CLI now writes an empty file and reports success. Before this PR that was an error. Treat empty content as "no snapshot yet".

voyti (commented):
Posting some new findings that seemed valuable

packages/cli/src/utils/run-in-cloud.ts
jamesbhobbs: Both fixed in 729053f: (1) synthesized no-op snapshots now report `artifactStatus: "synthesized"` instead of `"saved"` in `-o json`/`-o toon`, and human output prints an explicit note that an output-free snapshot was synthesized from the local source rather than produced by the run; (2) empty snapshot content is now treated as not-yet-attached inside `waitForRunSnapshot` — it keeps settling and returns null if the content stays empty, so the CLI reports `not_produced` and exits 1 instead of writing an empty file. Added tests for the synthesized status (JSON + human disclosure), empty-then-real content, and persistently-empty content. Please resolve if satisfied.

skills/deepnote/references/cli-run.md
jamesbhobbs: Fixed in b386cc7: `cli-run.md` and the CLI README now document the actual contract — `success` is the overall command outcome (run succeeded AND its snapshot was delivered), and a successful run with no snapshot exits 1 with `artifactStatus: not_produced` and an `artifactError`, including the `--notebook-id`-only case. They also cover the new `artifactStatus: synthesized` value from the sibling thread. You are right about the PR description too — a corrected version (success keeps the meaning it had on main; the real change is artifact failures moving from `error` into `artifactStatus`/`artifactError`) is written and will land on the description shortly. Please resolve if satisfied.

jamesbhobbs:
Chain approved on subsequent PR so this is approved by Wojtek

packages/cli/src/utils/run-in-cloud.ts
```diff
@@ -375,31 +424,35 @@ export async function runInDeepnoteCloud(path: string | undefined, options: RunC
       })
       snapshotPath = written.snapshotPath
       timestampedSnapshotPath = written.timestampedSnapshotPath
-    } else if (success) {
-      snapshotError = `Run ${finalRun.runId} completed but returned no snapshot content.`
+      artifactStatus = synthesized ? 'synthesized' : 'saved'
+    } else if (lastRetryError !== undefined) {
```
voyti, line 428: Leaving this finding, as it indicates potential logic error:  "stale lastRetryError misclassifies a confirmed "no snapshot" as unavailable If an early status re-fetch fails transiently but a later one succeeds and shows no snapshot attached, the CLI throws  the old error anyway and reports artifactStatus: 'unavailable' with an unrelated network message instead of not_produced."

jamesbhobbs: Confirmed — `lastRetryError` was never cleared, so an early transient re-fetch failure survived a later successful re-fetch that freshly confirmed no snapshot, and the CLI reported `unavailable` with the unrelated network message instead of `not_produced`. Fixed in #479: `waitForRunSnapshot` now clears the tracked failure on each successful status re-fetch and reports it as `SettledRunSnapshot.retryError` only when the *last* re-fetch failed (final observation stale); the CLI throws that instead of its own accumulated state. All-failed re-fetches still report `unavailable`. Regression tests cover both directions. Please resolve if satisfied.

PR #479, opened 2026-08-19
fix(cloud,cli): stale retry error no longer reclassifies a confirmed no-snapshot as unavailable

jamesbhobbs:
## Summary

Fixes a misclassification voyti found post-merge on #448 ([comment](https://github.com/deepnote/deepnote/pull/448#discussion_r3815879676)): during snapshot settling, the CLI accumulated `onRetryError` failures and threw the last one whenever content ended up null — even when a **later status re-fetch succeeded** and freshly confirmed that no snapshot was attached. A single early transient 503 therefore turned a confirmed `artifactStatus: 'not_produced'` into `'unavailable'` with an unrelated network message.

### Fix

The success signal only exists inside `waitForRunSnapshot`, so the tracking moves there:

- `waitForRunSnapshot` clears its tracked re-fetch failure on every successful status re-fetch and reports it as `SettledRunSnapshot.retryError` — set **only when the last re-fetch failed**, i.e. when the final "no snapshot" observation rests on stale data.
- The CLI throws `settled.retryError` instead of its own accumulated state; `onRetryError` remains a pure debug-logging hook.

Behavior matrix (null content):

| Settle re-fetches | Before | After |
|---|---|---|
| all succeed | `not_produced` | `not_produced` |
| early failure, later success | `unavailable` + stale network error | `not_produced` |
| all fail (or last fails) | `unavailable` | `unavailable` |

## Test plan

- [x] New cloud tests: `retryError` cleared by a later successful re-fetch; reported when the final re-fetch failed
- [x] New CLI regression test: transient 503 then successful confirmation → `not_produced`, `artifactError` without the network message (fails on `main`)
- [x] Existing all-failed-re-fetches → `unavailable` test still passes
- [x] Full repo suite (2,919 tests), typecheck, biome/prettier, cspell green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of temporary snapshot retrieval failures.
  * Successful follow-up checks now clear stale errors and correctly recognize when no snapshot was produced.
  * Final retrieval failures are reported accurately while cancellation errors continue to propagate.

* **Tests**
  * Added coverage for transient and persistent snapshot retrieval failures.

### DIFF
--- a/packages/cli/src/utils/run-in-cloud.test.ts
+++ b/packages/cli/src/utils/run-in-cloud.test.ts
@@ -654,6 +654,39 @@ describe('runInDeepnoteCloud — output and exit codes', () => {
     expect(process.exitCode).toBe(ExitCode.Error)
   })
 
+  it('reports not_produced when a later re-fetch confirms no snapshot after a transient failure', async () => {
+    // GET#1 is the poll (terminal, no snapshot); the first settle re-fetch (GET#2) fails
+    // transiently; the later re-fetches succeed and freshly confirm no snapshot was attached.
+    // The stale early error must not reclassify that confirmed no-snapshot as unavailable.
+    let getCount = 0
+    vi.spyOn(global, 'fetch').mockImplementation(async (_url, init) => {
+      const method = (init?.method as string) ?? 'GET'
+      if (method === 'POST') {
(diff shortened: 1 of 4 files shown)


jamesbhobbs:
Trivial, this merging w/o human review

PR #483, opened 2026-08-20
chore(repo): update TypeScript to 7.0.2

jamesbhobbs:
## Summary

- update TypeScript from 5.9.3 to 7.0.2
- update tsdown to 0.22.14 for TypeScript 7 compatibility
- remove the obsolete vite-tsconfig-paths plugin and use Vite 8 native tsconfig path resolution
- migrate path aliases away from the removed baseUrl compiler option

## Validation

- pnpm test (2,952 passed, 1 skipped)
- pnpm typecheck
- pnpm biome:check
- pnpm build
- pre-push lint/format and spell-check

## Summary by CodeRabbit

* **Chores**
  * Updated TypeScript and build tooling versions for improved compatibility.
  * Simplified development configuration and removed an obsolete development dependency.
* **Refactor**
  * Improved path resolution across the project.
  * Standardized package builds to support flexible file extensions across supported output formats.
  * Updated build configurations across multiple packages for more consistent artifact generation.


packages/blocks/package.json
```diff
@@ -27,7 +27,7 @@
     "package.json"
   ],
   "scripts": {
-    "build": "tsdown --format esm --format cjs --dts",
+    "build": "tsdown --format esm --format cjs --dts --no-fixed-extension",
```
dinohamzic, line 30: It seems this breaks watch mode, eg. 
- I run `pnpm --filter @deepnote/blocks watch` 
- the watcher creates `index.mjs`, but the package still tells node to load `index.js`
- node can't find that file so dependent packages crash with "module not found"

I guess we should add `--no-fixed-extension` to the watch command below as well?

packages/database-integrations/package.json
```diff
@@ -27,7 +27,7 @@
     "package.json"
   ],
   "scripts": {
-    "build": "tsdown --format esm --format cjs --dts",
+    "build": "tsdown --format esm --format cjs --dts --no-fixed-extension",
```
dinohamzic, line 30: Same as above

dinohamzic (changes requested):

packages/blocks/package.json
jamesbhobbs: Good catch — added `--no-fixed-extension` to the watch script as well. Pushed in 5c44a1884a. Mind resolving if that looks right?

packages/database-integrations/package.json
jamesbhobbs: Fixed here too — added `--no-fixed-extension` to the watch script. Mind resolving if that looks right?

dinohamzic (approved):

PR #496, opened 2026-08-27
feat(pipelines): compose Deepnote notebook runs into pipelines

jamesbhobbs:
**1 of 5.** Stacked: this → #497 (file format) → #498 (demo) → #504 (TS SDK) → #505 (Python SDK). Review in order; each is independently reviewable.

Run several existing Deepnote notebooks as one pipeline from plain TypeScript control flow: fan out with `Promise.all`, gate on the results, decide.

```ts
import { runPipeline } from '@deepnote/pipelines'

const { value, graph } = await runPipeline(async ({ run, control, outputs }) => {
  const analyses = await Promise.all(
    REGIONS.map((r) => run({ id: r.name, notebookId: r.notebookId, inputs: { region: r.name } })),
  )
  const readings = analyses.map((s) => outputs.lastJson(s))
  const failing = await control(
    { id: 'quality-gate', kind: 'gate', dependsOn: analyses.map((s) => s.id) },
    () => readings.filter((r) => r.qualityScore < 0.95).map((r) => r.region),
  )
  return { checked: readings.length, failing }
}, { token, onEvent })
```

Deliberately an imperative API, not a workflow language. The callback's own control flow provides sequencing, concurrency, and conditionals; the library records what happened as a graph, an event stream, and normalized step results.

### Where it lives

A new package, `@deepnote/pipelines`. Nothing reachable from `runPipeline` imports `node:*`, so the same pipeline runs in a script, in CI, and in a browser page (#498 proves that). A local, Python-backed runner was the wrong home for that, so the browser-safe half of `@deepnote/local-runner` (`snapshot-view`, `input-info`, output extraction) moved here and local-runner re-exports it. Nothing already published changes shape.

Notebooks are addressed by id and must already exist. Running a pipeline needs permission to run a notebook, not to create one, which is what lets a page do it with a viewer's short-lived token.

### API

- `runPipeline(workflow, { token, concurrency?, onEvent? })`: the pipeline API against Deepnote Cloud.
- `runPipelineWithExecutor(workflow, options, executor)`: same engine, runner left open. A caller with a local Python kernel supplies one here.
- `control()` records a local decision as a graph node, so a gate is visible instead of happening invisibly between steps.
- `pipelineOutputs.lastJson` / `lastAgentText` read results without depending on block ids, which Deepnote reassigns when it creates a notebook.
- A failed notebook rejects with `PipelineStepError` carrying the result; `allowFailure: true` returns it instead, for every way a step can fail including timeouts and API errors. Every rejection carries `partial`: the steps that finished, the graph, and the timings.
- `concurrency` (default 10) caps steps in flight, so a wide fan-out cannot flood the API.

- **`allowFailure` now covers a throwing executor.** `executeNode` only honoured it when the executor returned `success: false`; a poll timeout, a transport or API error, or a missing snapshot was wrapped in `PipelineStepError` and rethrown, so one hung run in an `allow_failure` fan-out discarded every other step (seen live on 2 September: `onboarding-brief[0]` timed out after ten minutes and five finished steps were lost). The engine now synthesizes the failed result the executor could not return: `success: false`, `status: 'timeout'` for a `RunTimeoutError` and `'error'` otherwise, the message as `error`, empty outputs, no snapshot, the `runId` when the error exposes one, and the timings. The graph node finishes as failed and `step_failed` carries that result. Without `allowFailure` nothing changes.
- **A failure no longer discards the run.** `runPipelineWithExecutor` awaited the callback with no `try/catch`, so a failing step lost the results and graph it had accumulated. `PipelineStepError` gains `partial` (`steps` in start order, `graph`, `startedAt`, `finishedAt`, `durationMs`), and anything else the callback throws is wrapped in a new exported `PipelineRunError` with the same `partial` and the original as `cause`. The resolved shape is unchanged. The example runner writes whatever partial result it receives on failure.
- **The whole last output no longer has to be JSON.** Jupyter merges consecutive prints into one stream chunk, so a summary line before the JSON made `lastJson` report no structured output. `lastOutputJson` and `outputJson` now fall back to the suffix starting at the last line beginning with `{` or `[`, scanning backwards so pretty-printed JSON parses too. When nothing parses, the error quotes the last 200 characters the block printed.
- README: a "When a step fails" section, and `lastJson` documented as needing the last block's output to end with JSON, with earlier lines ignored.

### Testing

Engine covered through a fake executor (graph, edges, control nodes, concurrency cap with an exact peak-in-flight assertion, duplicate/unknown id rejection, output helpers); the cloud binding with `@deepnote/cloud` mocked; abort-during-sleep in `packages/cloud`. Full suite (3135), typecheck, biome, prettier, cspell green.

### Notes for review

- Replaces #435, which mixed four concerns and was built server-first. This stack is what it looks like built client-only from the start.
- `target` on a result is a string named by the executor, so a local-kernel executor later needs no type change.

## Summary by CodeRabbit

- **New Features**
  - Added pipeline execution for concurrent notebook workflows, control nodes, result gates, graph metadata, and configurable concurrency limits.
  - Added cloud-backed execution with input validation, status updates, output extraction, diagnostics, and cancellation support.
  - Added input metadata helpers for sliders and select controls.
  - Added partial-result reporting when pipeline steps fail.
  - Added a script-based regional pipeline example with quality checks.

- **Documentation**
  - Added installation, usage, browser compatibility, execution, and configuration guidance.

- **Bug Fixes**
  - Pipeline polling now stops promptly when cancellation is requested.
  - Improved JSON output parsing and failure details.


jamesbhobbs:

PR #497, opened 2026-08-27
feat(pipelines): define a pipeline in a .deepnote file

jamesbhobbs:
**2 of 5.** Stacked on #496. The diff shown here is against #496's branch.

A pipeline can live in a file instead of in code: a notebook marked `isPipeline: true` whose `notebook-function` blocks each name a module notebook, its inputs, and the values it publishes. The encoding is the one deepnote.com already stores for module imports, extended with four typed keys, so the same file can later be executed natively by the product against the same planner.

```yaml
notebooks:
  - id: sales-pipeline
    name: Sales pipeline
    isPipeline: true
    blocks:
      - id: analyze-europe
        type: notebook-function
        metadata:
          function_notebook_id: nb-regional
          function_notebook_inputs:
            region: { custom_value: Europe }
          function_notebook_export_mappings:
            result: { enabled: true, variable_name: europe }

      - id: recover
        type: notebook-function
        metadata:
          function_notebook_id: nb-regional
          function_notebook_for_each: belowThreshold
          function_notebook_for_each_as: region
          function_notebook_inputs:
            region: { variable_name: region }
          function_notebook_export_mappings:
            result: { enabled: true, variable_name: recovered }

      - id: aggregate
        type: notebook-function
        metadata:
          function_notebook_id: nb-aggregate
          function_notebook_run_if: europe.qualityScore >= 0.95 || recovered != null
          function_notebook_inputs:
            europe_json: { variable_name: recovered, fallback: { variable_name: europe } }
```

### Dependencies come from variable flow

An input with `variable_name` references a value another step exports through `function_notebook_export_mappings`. That is the dependency edge; nothing is declared twice, and independent steps are independent by construction. Identifiers in `run_if`, the `for_each` source, and every `fallback` alternative are edges too.

### The primitives

| Key | Meaning |
|---|---|
| `function_notebook_run_if` | Gate a step on earlier results. Evaluated per element on a fan-out. |
| `function_notebook_for_each` / `_as` | One concurrent run per element of an exported array. Capped at 50. |
| `fallback` on an input | First alternative with a value, so a skipped producer does not cascade a skip. |
| `function_notebook_allow_failure` | Return a failed step result instead of failing the pipeline. |

A step whose inputs never arrived is skipped, including a `for_each` whose source was skipped. A failed step without `allow_failure` fails the pipeline; unrelated steps in flight finish. The rejection carries the engine's `partial` run plus the runner's `variables`, `skipped` and `failed` so far, so a caller can render what finished.

### Interpreted, not executed

The parent is read as a manifest. Deepnote's block engine runs blocks strictly in order, so executing the parent would serialize the fan-out into one run with one status. `planPipeline(file)` returns the graph without running anything. Plan-time errors cover everything checkable without running: missing or duplicate `isPipeline`, unknown references, duplicate exports, a step naming no notebook, `for_each_as` without `for_each` or shadowing an export, malformed conditions, cycles.

### Deliberately not JavaScript

A pipeline definition is data. The condition language has no `eval`, no calls, no assignment, own-property lookups only. `europe.__proto__` evaluates to false and `fetch("x")` fails at plan time. One rule for every comparison operator: numeric when both operands are numbers or numeric strings, strict otherwise, because the runs API delivers inputs as strings.

### Conformance fixtures

`test-fixtures/pipeline-conformance/` holds `.deepnote` files with expected plans (`linear`, `condition-only-dependency`, `fan-out-and-gate`) and one with expected execution results (`skip-and-fallback`). These are the language-neutral spec a native implementation must match.

- **Docs and a test for `allow_failure` on a timed-out run.** The engine fix on #496 makes `function_notebook_allow_failure` cover a throwing executor, so a fan-out with one hung element now collects from the rest and continues; a test here proves it against the file runner (two-element list published, the step in `failed`, the downstream step reached, `step_failed` carrying the synthesized `timeout` result). `blocks-pipeline.md` and the README say `allow_failure` covers every way a step can fail, that the result's `status` and `error` say which, and that a timed-out run may still be executing in Deepnote with its `runId` on the result. No engine change on this branch.
- **Unreadable exports are failed steps.** `exportsFrom` threw a plain `Error` from inside `publish` when a step's output held no JSON object or lacked an exported key, and `allow_failure` did not cover it because `publish` only checked `result.success`. With `allow_failure` the step is now listed in `failed`, publishes nothing and emits nothing new (a fan-out collects only its readable elements); without it the pipeline fails with a `PipelineStepError` naming the step.
- **The file runner's state rides on the error.** `runPipelineFile` and `runPipelineFileWithExecutor` attach `variables`, `skipped` and `failed` alongside the engine's `partial` (new exported `PipelineFileError` type). A terminal `allow_failure` step therefore lets the run resolve with every other variable present.
- **Documented input constraint.** The runs API rejects arrays for text inputs, so a variable holding a fan-out's collected list can only feed an `input-select` block with `deepnote_allow_multiple_values: true`. Stated in `blocks-pipeline.md` with an example; no engine change.
- Docs now say exports are read from the JSON object that ends the last block's output, with anything printed on earlier lines ignored, and spell out the failure semantics above in the README and `blocks-pipeline.md`.
- Rebuilt onto the new #496 tip as one squashed commit plus the two commits above.

### Open items

- Module-target validation is not possible yet: `getNotebook` in `@deepnote/cloud` does not expose an `isModule` flag. Worth adding to the v2 notebook response.
- A variable referenced only in `run_if` whose producer was skipped reads as absent rather than skipping the step, so `recovered == null` works as a gate. Confirm this is the semantics the product wants.
- `toRunInputs` still rejects object inputs, so structured values flow between steps as JSON strings.

### Testing

Planner, scheduler, condition language, schema round-trip, and conformance fixtures, including a peak-concurrency assertion that sequential execution could not pass. Full suite (3218), typecheck, biome, prettier, cspell green.

## Summary by CodeRabbit

* **New Features**
  * Added support for defining, planning, and running pipelines directly from `.deepnote` files.
  * Added conditional execution, dependency-based scheduling, fallback inputs, failure tolerance, and dynamic fan-out.
  * Added configurable concurrency controls for pipeline steps.
  * Added typed notebook-function inputs, exports, pipeline metadata, and public planning and execution APIs.
  * Added support for partial results, skipped steps, and failed-step reporting.

* **Documentation**
  * Added guidance and examples for file-based pipelines, conditions, fan-out, dependencies, and execution behavior.

### DIFF
--- a/examples/pipelines/sales-pipeline.deepnote
+++ b/examples/pipelines/sales-pipeline.deepnote
@@ -1,0 +1,204 @@
+# A pipeline defined as a file, not as application code.
+#
+# The notebook marked `isPipeline: true` is the pipeline. Each of its notebook-function blocks is one
+# step: the notebook it runs, the inputs to run it with, and the values it publishes. This is the
+# encoding deepnote.com stores for a notebook-function block — an input is `{ variable_name }` (a
+# pipeline variable) or `{ custom_value }` (a literal) — so the same file means the same thing in the
+# product and here.
+#
+# Dependencies are not declared twice: a step that reads `{ variable_name: portfolio }` depends on
+# whichever step exports `portfolio`, so the three regional analyses below are independent by
+# construction and run concurrently.
+#
+# The pipeline notebook is read as a manifest rather than executed. Running it through Deepnote's
+# block engine would serialize these steps into one run with one status; interpreting it keeps the
+# concurrency and reports each step separately.
+#
+# Each notebook ends by printing one JSON object; the export mappings read fields from it. Deepnote
+# input blocks take strings, booleans and lists of strings, so a structure that has to cross into
+# another notebook is exported as a JSON string (the `*_json` fields below).
+#
+# Replace every function_notebook_id with the id of the notebook in your Deepnote project.
+metadata:
+  createdAt: '2026-08-27T00:00:00.000Z'
+project:
+  id: local-runner-sales-pipeline
+  name: Pipelines demo - sales review
+  notebooks:
+    - id: local-runner-sales-pipeline-notebook
+      name: Sales review pipeline
+      isPipeline: true
+      blocks:
+        - blockGroup: analyze-north-america-group
+          id: analyze-north-america
+          sortingKey: a0
+          type: notebook-function
+          metadata:
+            name: North America analysis
+            function_notebook_id: REPLACE_WITH_REGIONAL_NOTEBOOK_ID
+            function_notebook_inputs:
+              region:
+                custom_value: North America
+              trailing_months:
+                custom_value: '6'
+              monthly_target_k:
+                custom_value: '420'
+            function_notebook_export_mappings:
+              summary_json:
+                enabled: true
+                variable_name: northAmerica
+        - blockGroup: analyze-europe-group
+          id: analyze-europe
+          sortingKey: a1
+          type: notebook-function
+          metadata:
+            name: Europe analysis
+            function_notebook_id: REPLACE_WITH_REGIONAL_NOTEBOOK_ID
+            function_notebook_inputs:
+              region:
+                custom_value: Europe
+              trailing_months:
+                custom_value: '6'
+              monthly_target_k:
+                custom_value: '340'
+            function_notebook_export_mappings:
+              summary_json:
+                enabled: true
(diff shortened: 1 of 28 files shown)


jamesbhobbs:

PR #498, opened 2026-08-27
feat(examples): a client-only pipeline demo

jamesbhobbs:
**3 of 5.** Stacked on #497 → #496. The diff shown here is against #497's branch.

`examples/local-runner/run-app` runs entirely in the browser and is a demo of one thing: every capability it uses is a library call. `serve.mjs` is a static file host with **no API routes**; publish the folder and nothing of ours runs.

| What the page does | Library call it makes |
|---|---|
| notebook and input definitions | `getNotebook` |
| run a notebook | `triggerNotebookRun` → `pollRunUntilComplete` → `waitForRunSnapshot` |
| show a past run | `getRun` + `fetchSnapshotContent` |
| run history | `listNotebookRuns` |
| schedule | `upsertNotebookSchedule` |
| orchestrated pipeline | `runPipeline` from #496 |
| provider decisions | `pipelineOutputs.lastJson` |

There is no `fetch(` left in the page. All of the above reaches the page through `@deepnote/pipelines/browser`, one IIFE bundle (`browser.iife.js`, global `DeepnotePipelines`) that runs in a bare `node:vm` context with no `process`, no `require`, and no Node globals.

### Auth

A short-lived, project- and viewer-scoped token from the Deepnote shell over `postMessage`. The token and the API origin arrive as one bundle and are used as one. The shell origin is fixed in `APP_CONFIG` and cannot be overridden by a query parameter. `?token=` exists for local testing only.

Nothing is created on the fly: notebooks are addressed by id and must already exist. The `.deepnote` files in the folder are the *source* for notebooks you create once in Deepnote.

### Decisions as data

The GPT, Claude, and arbiter notebooks end with a code block that emits one JSON line (`decision`, `rationale`, `nextAction`). The page reads it with `lastJson` instead of matching prose with a regex, so a model that phrases its verdict differently cannot break the demo.

- Every hand-rolled `fetch` replaced with the corresponding `@deepnote/cloud` function, exported through the browser bundle.
- Decision parsing moved from regex over agent text to a JSON output.
- Dead local-run branches, the `?shellOrigin=` override, and the duplicate `examples/local-runner/cloud-app` removed.
- README rewritten around the actual bundle exports; stale function names fixed.

### Known limits

- **Run history depends on token scope.** A published app's token may not enumerate runs; the panel hides itself rather than erroring.
- **No local Python kernel.** Every run goes through the API.
- Two things the library types do not yet carry are read off the raw run response, which the schema passes through: the run's `viewUrl` and the `snapshotBlocks` a static-app token receives instead of a raw snapshot. Worth adding to `NormalizedRun`.

### Testing

The page's API layer is extracted into a `vm` sandbox and exercised against the real library functions with `fetch` stubbed, so the assertions are about what the page sends (URL, bearer header, request body) and how it renders what comes back. Full suite, typecheck, biome, prettier, cspell green.

## Summary by CodeRabbit

* **New Features**
  * Added a browser-based local runner for executing, scheduling, and reviewing Deepnote Cloud notebook runs.
  * Added regional analysis and multi-provider decision-review pipelines with quality checks, recovery, aggregation, arbitration, and structured results.
  * Added run history, live progress, snapshot handling, authentication, scheduling, and direct-run configuration.
  * Added a browser-ready pipelines package for cloud execution and scheduling APIs.

* **Documentation**
  * Updated local-runner documentation with browser-only setup, pipeline behavior, scheduling, publishing, and configuration guidance.

* **Removed**
  * Removed the cloud-app example, local execution guidance, and separate orchestration browser bundle.


jamesbhobbs:

jamesbhobbs:
@coderabbitai review

jamesbhobbs:
@coderabbitai review

jamesbhobbs:
@coderabbitai review

jamesbhobbs:
@coderabbitai review

jamesbhobbs:
@coderabbitai review

jamesbhobbs:
@coderabbitai review

jamesbhobbs:
@coderabbitai review

PR #500, opened 2026-08-27
feat(pipelines): run a .deepnote pipeline on a schedule

jamesbhobbs:
**5 of 5.** Stacked on #499 → #498 → #497 → #496.

Deepnote already schedules one notebook. Point that at a notebook that *interprets* a manifest and a whole pipeline runs on a schedule — the run is an ordinary Deepnote run (durable, retryable, visible in Deepnote's own UI) while the definition stays in a file. No workflow engine, no server of ours, no Temporal.

### Why not just schedule the manifest?

Because Deepnote's engine would do the wrong thing *quietly*, which is worse than failing:

- It **runs blocks strictly in order**, so a fan-out that should be concurrent is serialized.
- It **knows nothing about `run_if`, `for_each`, or `{{ }}`** — a conditional recovery step would run unconditionally, a fan-out would run once, and `{{portfolio}}` would arrive as that literal string.
- Native `notebook-function` inputs are **baked as a static JSON literal** at authoring time (`inputs=${JSON.stringify(inputs)}` in the codegen), so no value can flow from one step into the next.

So the scheduled artifact is a runner, not the definition.

### What ships

- `packages/local-runner/python/deepnote_pipeline.py` — planner, condition and reference languages, and a concurrent executor over the runs API. Stdlib + PyYAML.
- `examples/local-runner/scheduled-pipeline/runner.deepnote` — a self-contained notebook that embeds the interpreter, so it does not depend on this repo being in the project. Create it once and schedule it.
- `--plan` prints the DAG and runs nothing, which is the fastest way to check a manifest.

I verified the embedded cell actually executes and can plan a fixture from inside the notebook. That caught a real bug: a notebook cell runs with `__name__ == "__main__"`, so the CLI entry point had to be stripped when embedding or argparse would abort the cell.

### The thing I'd most want reviewed

**The semantics now exist in two languages.** That is a standing risk of quiet divergence, and it is the main cost of this PR.

The mitigation is `test-fixtures/pipeline-conformance`: both planners must produce identical plans for every fixture, enforced by `pipeline-conformance.test.ts`. Change `run_if`, `for_each`, `{{ }}`, or dependency derivation in one language and that test fails until it is changed in the other.

I checked the test actually has teeth by breaking the Python planner on purpose — **and the first fixture set did not catch it**, because every condition variable also appeared in inputs, so conditions never contributed a unique dependency edge. `condition-only-dependency.deepnote` exists because of that experiment, and with it the drift fails as it should. Worth keeping that property in mind when adding fixtures.

Python is skipped rather than failed when unavailable (`describe.skipIf`), since it is not needed to build or use the package — so this needs `python3` + PyYAML on CI runners to actually enforce anything. Worth confirming.

### Limits, stated in the example README

Steps run concurrently and the run is durable, but there is **no resume**: a failed run restarts from the beginning, and notebook runs are not automatically idempotent. For replay, per-step retries, and timers, #499's Workflow SDK integration is the right tool. These two are complementary — this one trades resume for needing no server at all.

## Summary by CodeRabbit

* **New Features**
  * Added a standalone Python pipeline interpreter with planning and execution modes.
  * Added scheduled pipeline runner support, including dependency resolution, conditions, fan-out execution, retries, and output handling.
  * Added a command to run the sales pipeline example.
  * Added cross-language pipeline conformance validation.

* **Documentation**
  * Added guidance for scheduling and running pipeline manifests, including setup, limitations, and examples.

* **Tests**
  * Added coverage for linear workflows, conditional dependencies, fan-out execution, recovery, aggregation, and gating.


jamesbhobbs:

jamesbhobbs:
Closing in favour of native scheduled execution in deepnote.com. Interpreting a manifest through a Python reimplementation embedded in a notebook creates a second implementation of the planner semantics, and it already diverges from the TypeScript one at execution time (barrier scheduling between waves, failure surfacing, comparison coercion). Two things from this PR carry forward: the conformance fixtures move into #497, re-encoded to the native input shape, and the "why not schedule the manifest directly" analysis (strict block ordering, no `run_if`/`for_each`, inputs baked at codegen) is the requirements list for the native executor.

PR #507, opened 2026-09-01
chore: ignore local worktrees under .claude/worktrees

jamesbhobbs:
Agent worktrees are created inside the repo at `.claude/worktrees/`, where git sees them as untracked gitlinks — they can be (and were) accidentally staged. Adding the directory to `.gitignore`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #508, opened 2026-09-01
fix(cli,skill): let sync and publish share one baseline, and add an app-models reference

jamesbhobbs:
Two strands, both about the same confusion: which command owns a project's files, and which app model a given deliverable actually is.

1. **`packages/`** — sync and publish stop overwriting each other's writes to the project file store.
2. **`skills/`** — a canonical reference for Deepnote's app models, and a corrected skill sync map in `AGENTS.md`.
3. **`deepnote static-site access`** — change a published site's sharing and viewer API access without republishing (section at the end).

---

# 1. Sync and publish share one baseline

## The conflict

`deepnote publish` deploys into `_deepnote_static/**`, which is a subtree of the project file store that `deepnote sync --all-files` mirrors. Neither command knew the other existed — `STATIC_ROOT` was a private const in `publish.ts`, and sync claimed the entire inventory. Four consequences:

1. **Guaranteed churn.** Publish never touched sync's manifest, so after every publish the whole static subtree looked changed and the next `sync --all-files` re-downloaded the built site into `.files/`.
2. **Push could regress a live site.** `uploadProjectFiles` never fetched the inventory at all — it was documented "last-write-wins, no staleness check", with no equivalent of the `baseModifiedAt`/`baseContentHash` protection the notebook path gets.
3. **`--prune` produced ghosts.** `publish --prune` removed a remote asset; sync's local copy still hash-matched the manifest so push skipped it, leaving it on disk forever and resurrecting it in the cloud on the next edit.
4. **Round-trip loop.** `dist/` → publish → cloud → sync pull → `.files/` → git → sync push → cloud.

## The fix

Coordination, not partition — both commands keep writing the same paths.

**Sync gains per-file lost-update protection on push.** Every candidate is now checked against the cloud inventory, and a file that moved since the manifest baseline goes through the existing `--on-conflict` override-or-skip choice instead of being overwritten. Reported as `N file(s) kept from Deepnote` / `filesSkipped`. This is the load-bearing part: publish often runs from CI where no manifest exists, and the Deepnote app can write these paths too. It also fixes the same silent overwrite for ordinary working files (`requirements.txt` edited in the app was clobbered before).

**Publish updates the sync mirror** when the published directory sits inside a synced workspace: it writes each file into the project's `.files/` mirror and records size, hash, and server `updatedAt`, exactly as a sync download would. Manifest, mirror, and cloud then agree, so sync sees the deploy as already in step. `--prune` drops pruned paths from both. `--sync-root <dir>` / `--no-sync-root` control discovery, which otherwise walks up from the published directory.

**Publish stops before mutating anything** if a path it would write has moved on in Deepnote since that workspace last synced — the mirror holds no copy of that content. `--force` overrides. Its check is deliberately narrower than sync's: sync is a mirror where both sides are authoritative, publish is a deploy where the local build is authoritative, so a path with *no* baseline is not flagged (that's the normal state of a static root written by earlier publishes, and flagging it would break the first publish in every workspace).

`PROJECT_STATIC_ROOT` moves to `@deepnote/cloud` so every writer agrees on where the boundary is.

## Edge cases handled

- A **pending replacement** is exempt from the conflict check — that missing cloud copy is sync's own unfinished delete, not another writer's deletion. Without this, every interrupted upload would become a false conflict.
- The mirror is only updated when the **tracked project directory already exists**. Creating it would flip `readLocalNotebookFiles` from `null` to `[]`, which sync reads as "every notebook was deleted locally" and pushes. Leaving the baseline stale is safe: the next sync pulls once and converges.
- A record written **before `updatedAt` was tracked** has no baseline to compare, so it still overwrites (rather than manufacturing conflicts for existing users) and becomes verifiable after the next pull.
- **Mirror failures are warnings, not errors** — the deploy already succeeded, and a stale manifest is safe because the next sync brings the mirror back in step (a pull re-downloads the published files, a push surfaces them as a conflict).
- Push hashes files in one pass and re-reads them at upload time, so one prompt can cover a whole project without buffering every file in memory.

## Known limitation

The file API has no conditional write (`POST /v2/files` refuses to overwrite at all, which is why writes are delete-then-upload), so this is optimistic concurrency over the inventory: a small time-of-check/time-of-use window remains, and each replaced file has a brief 404 window on the live site. Both need a server-side atomic replace to close, tracked separately.

## Review rounds (takeover, 2026-09-03)

Live-tested against the real CLI with a mock server (`--contract strict|echo`) and re-reviewed in five rounds. Changes on top of the original design:

- **Server dependency made explicit.** The per-file `updatedAt` baseline that publish records comes from the upload response, which the server only started echoing in deepnote-internal#20773 (in production since v0.16.997, 2026-09-03). On servers without it — self-hosted or older — publish-written baselines are unverifiable; the docs say so, and the CLI degrades to the pre-baseline overwrite rather than failing.
- **The publish stop needs a usable baseline.** Only an entry carrying the server's `updatedAt` is compared: `--all-files` syncs always record one, publishes do since #20773. A path without one is overwritten like any deploy. A per-path warning for that case was tried and removed — its only remedy, a pull, records the cloud copy as the baseline so the next publish passes the stop anyway.
- **Pull keeps the baseline of a cloud-deleted file that stays on disk** (no `--prune`), and warns. Previously the record was dropped, so an edited copy silently resurrected a file someone had removed (e.g. via `publish --prune`). An edited push now surfaces the deletion as a conflict; `--prune` still removes file and record together.
- **Pending uploads.** Publish clears a path from `pendingFileUploads` when it writes or prunes it, so an interrupted sync replacement that publish already settled cannot brick later syncs. A pending path whose cloud copy exists again is a conflict ("re-created in Deepnote after an interrupted upload"), not a silent retry — the "our own unfinished delete" premise is disproven when the copy is back. Sync persists each settled upload immediately, so a run interrupted later cannot leave its own upload looking pending; if a kept re-created conflict does occur, skipping it drops the retry so the next pull resolves it instead of the conflict sticking.
- **Explicit `--sync-root` failures are exit 2 consistently**, including a tracked project directory that is missing; a manifest that exists but cannot be read (or is a symlink) also exits 2 with a pointer at `--no-sync-root`.
- **Kept-files warning is honest about pull**: it says pulling replaces the local copies; the download-side dirty check itself is a follow-up.
- **Token handling unified**: every command resolves `--token`/`DEEPNOTE_TOKEN` through one trimming helper (a pasted token with a stray space used to fail only in `publish`).
- **Dry run honours an explicit `--on-conflict`** instead of always reporting a skip; the upload loop re-checks the 100 MiB cap on the bytes it actually uploads.
- Shared `sha256`/`baselineDiverged` helpers so publish and sync cannot drift on what "changed" means; `apps.md` states the static site's URL-to-file mapping.

Gates at head: `pnpm test` 3147 passed, typecheck, biome, prettier, spell-check clean; CI green including codecov patch/project.

---

# 2. App models reference (#509)

The skill had no canonical account of Deepnote's app models, so an agent choosing between them had to infer the differences from `cli-publish.md`, two example READMEs, and the local-runner package docs — and the one place that mentioned app tokens covered only the CLI flag.

**`skills/deepnote/references/apps.md`** states the five models, what each is made of, where it runs, and which credentials it gets: data apps, Streamlit apps, published static sites, published browser apps with viewer API access, and local Node-backed `serveStatic` apps. It leads with a decision table keyed on the questions that actually change the plan — including whether an agent can create the underlying files at all, which is where hosted-only workflows quietly fail.

It also records the failure mode worth knowing before it bites: code developed against a local preview with a personal token keeps working there and does nothing once embedded, with no error, because the viewer token is limited to one run loop.

**`AGENTS.md`** — the skill sync map claimed MCP tools mirror the CLI and should be recorded in `cli-*.md`. They do not: `@deepnote/mcp` (local files), hosted MCP at `deepnote.com/mcp`, and the codex-plugin consumer are three separate surfaces. The map now routes each to its own home and adds a row for app models, publishing ownership, and token boundaries.

**`skills/deepnote/SKILL.md`** — an Apps section pointing at the new reference.

## Ownership documented consistently

Both strands land the same rule, which is why they read better together than apart. `apps.md` states it for agents and the `cli-*.md` references carry the mechanics:

- `deepnote publish` is the deploying writer for `_deepnote_static/**` — build output goes through publish, not through a synced workspace.
- The two commands share one baseline rather than splitting the namespace: sync mirrors those paths and surfaces a republished site as a conflict instead of reverting it.
- Publish updates a surrounding workspace's mirror unless told not to (`--no-sync-root`, which is what CI wants), and exits 1 without touching the project when Deepnote holds unsynced changes.

A follow-up commit on this branch dropped an earlier "publish owns the namespace" line from `apps.md`, which read as a split now that the implementation deliberately shares a baseline.

---

## Branch history

- **#509** was merged into this branch (its base was `worktree-sync-publish-coordination`), bringing `apps.md`, the `AGENTS.md` map fix, and the `SKILL.md` pointer.
- **`main` was merged in** to pick up #507. That is why #507 appears in this branch's log but not in the diff — it is already on `main`.
- One follow-up commit states the publish/sync coordination in `apps.md` now that the mechanism ships in this PR.

## Testing

The merged commits touch only `AGENTS.md`, `SKILL.md`, and `skills/deepnote/references/apps.md` — no code, so the gate results for the implementation still stand: `pnpm test` (3127 passed, 1 skipped), `pnpm typecheck`, `pnpm biome:check`, `pnpm prettier:check`, `pnpm spell-check` all clean.

New test coverage in this PR:

- `publish.test.ts` — mirror writes and manifest records, the divergence stop and `--force`, `--prune` clearing the mirror, `--no-sync-root`, both `--sync-root` error cases, the missing-project-directory case, and the warn-not-fail mirror failure.
- `sync.test.ts` — a diverged cloud copy kept vs. overridden, a deleted cloud copy treated as a conflict, an unchanged copy not flagged, a pending retry finishing when the cloud copy is gone, plus unit coverage of `describeCloudFileDivergence`.
- `publish-mirror.test.ts` — `findDivergedPublishPaths` (including the deliberate mirror/deploy asymmetry) and `resolvePublishMirror` discovery.

Docs updated per `AGENTS.md`: `packages/cli/README.md`, `skills/deepnote/references/cli-publish.md`, `skills/deepnote/references/cli-sync.md`, and both commands' `--help`.

## Stacked on this

**#510** adds the public `docs/` pages for `deepnote publish` and `deepnote sync`, and distinguishes the CLI `deepnote sync` command from the in-product "Deepnote file sync" feature. It is based on this branch and will retarget to `main` when this merges.

## Summary by CodeRabbit

* **New Features**
  * Added `static-site access` to manage published-site sharing and viewer API access without republishing.
  * `publish` now supports sync-workspace coordination, conflict detection, pruning, and force overrides.
  * Sync now detects changed, deleted, or recreated cloud files and reports uploaded and skipped files more clearly.

* **Bug Fixes**
  * Prevented publishing and syncing from silently overwriting newer or deleted cloud files.
  * Improved validation for workspaces, paths, authentication, and mirror updates.

* **Documentation**
  * Expanded guidance for apps, publishing, syncing, access controls, and MCP surfaces.

---

## Static-site access lifecycle

This draft now also adds `deepnote static-site access`, a settings-only command that reuses the existing `PATCH /v2/projects/{projectId}` contract. It can disable or re-enable sharing and change viewer API access without uploading or deleting `_deepnote_static/**`. Disabling sharing also disables viewer API access; re-enabling serves the retained files again.

The CLI README, `cli-publish.md`, `apps.md`, and the installed Deepnote skill describe the same publish → change access lifecycle. Six command tests cover combined settings, sharing-only and API-only updates, invalid empty/contradictory requests, and API failures.

Latest local verification: `pnpm typecheck`, `pnpm biome:check`, `pnpm prettier:check`, `pnpm spell-check`, and `pnpm test` (3133 passed, 1 skipped).


skills/deepnote/references/apps.md
```diff
@@ -0,0 +1,164 @@
...
+
+Browser files — HTML, JS, CSS, assets — stored below `_deepnote_static/**` in the project's file
+store and served by Deepnote. There is no server: whatever the browser can do, the app can do.
+
+```bash
+deepnote publish ./dist --project-id <uuid>
+```
+
+`deepnote publish` deploys into `_deepnote_static/**`: it replaces matching files, optionally prunes
+stale ones, enables website sharing, and prints the canonical URL. Use that returned URL — never
```
tkislan, line 58: @voyti we should mention here explicitly, that it works like a regular webserver, where index.html is a root path, and any other html file basename transforms to URL path

voyti:
Live test + static review round findings (agent summarized), I'll handle this scope:

Finding 1 — the server never returns updatedAt from POST /v2/files, and most of the PR's value hangs on it (blocking,  already essentially done as a separate small PR on the server side - internal 20773)

When the CLI uploads a file, it asks the server "when was this file last changed?" and writes the answer down in the sync manifest. Later, it compares that note against the server to decide "has anything changed since I last looked?" The problem: the real Deepnote server, when you upload a file, answers only with "here's the file's path" — no timestamp at all. So every note the CLI writes after a publish or a push has a blank where the timestamp should be.

Finding 2 — --prune can brick every subsequent sync of the project (small code fix, will add that)

When a sync upload gets interrupted halfway, the CLI leaves itself a reminder: "finish uploading this file next time." If you then run publish --prune and it deletes that same file (from the cloud and from your local mirror), the reminder is never crossed out. Next sync, the CLI reads the reminder, looks for the local file, finds it gone, and refuses to continue — and it refuses again on every sync after that, until you hand-edit the manifest JSON.

Finding 3 — the conflict message steers users into an unprotected overwrite (will fix wording; the gap itself can be a potential follow-up)

The PR adds a safety prompt on push: "these files changed in Deepnote, keep the Deepnote copy or overwrite?" If you choose to keep the Deepnote copy, the message advises "Pull to bring them down before pushing again." But pull has no such safety check — it overwrites your local files without asking. So a user who just carefully said "don't destroy anything" and then follows the tool's own advice loses their local edits, silently, with no copy anywhere.

Finding 4 - (one-line check: a pending path that does exist in the cloud is a conflict, ask like the others)
The exemption at [sync.ts:571](https://github.com/deepnote/deepnote/blob/7e0c316/packages/cli/src/commands/sync.ts#L571) rests on the premise "a missing cloud copy is our own unfinished delete," and when the cloud copy exists, that premise is false

A file with a pending "finish this upload" reminder skips the safety check, on the assumption its missing cloud copy is sync's own unfinished delete. But if someone else has since put a new file at that path, sync deletes and overwrites it with no prompt — under default settings.

Three smaller items I'll cover:
- Token with a stray space: works in some commands, fails cryptically in publish — five commands clean the token four different ways. Fix: use the existing normalizeToken helper everywhere.
- Explicit --sync-root whose project folder is gone: currently succeeds silently without doing what you asked. Should fail with exit 2, like the other "you explicitly asked, I can't deliver" cases. Automatic discovery keeps its silent skip.
- Corrupt sync manifest above the build folder blocks publishing: correct fail-safe, keep it — but the error should mention the --no-sync-root escape hatch and use exit 2, not 1.

+Also will cover Tomas' remark about index.html and webserver behavior

skills/deepnote/references/apps.md
voyti: Added a note of that

packages/cli/src/commands/sync.ts
```diff
@@ -535,20 +565,75 @@ async function uploadProjectFiles(
...
+  }
+
+  for (const { relPath, conflict } of planned) {
+    if (conflict !== undefined && !overrideConflicts) {
+      // A kept re-created cloud copy is no longer ours to finish replacing. Dropping the retry turns
+      // the path back into an ordinary diverged file: pull brings the cloud copy down, push asks again.
+      if (!ctx.dryRun && pending.delete(relPath)) {
+        commitPending()
+        await persistManifest()
+      }
```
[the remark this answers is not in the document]
voyti: checked the internal DELETE /v2/files handler and the contracts: neither delete nor POST /v2/files accepts any precondition. The CLI has nothing to pass an expected updatedAt into, so the requested fix is impossible from the OSS side. It's a server feature, the same shape as #20773 was.

skills/deepnote/references/cli-publish.md
```diff
@@ -36,6 +39,62 @@ API access is security-sensitive and is not enabled by default. Pass `--api-acce
...
+re-downloading the whole site. A server that does not echo `updatedAt` on upload leaves those
+baselines unverifiable until the next pull — the divergence stop does not cover such paths, and the
+next pull re-downloads them once. `--prune` also removes the pruned paths from the mirror and manifest,
+so a later push cannot resurrect them.
+
+Before writing anything, publish compares the inventory it already fetched against the manifest
+baseline. If a path it is about to write or prune has moved on in Deepnote since that workspace last
+synced, publish exits 1 without mutating the project — the mirror holds no copy of that content.
+Run `deepnote sync --all-files` to bring it down, or pass `--force` to overwrite. A path deleted in
+Deepnote is not a stop: the deploy re-creates it from the build, and nothing is lost. Only an entry
```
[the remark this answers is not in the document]
voyti: this was already declined as a server gap needing a conditional-replace endpoint

voyti:
## Bugs found outside this PR

While end-to-end testing this branch I ran into a handful of issues that are **not introduced by this PR** — pre-existing CLI behaviour, server-side properties, or unrelated repo debt. Recording them here for visibility, not as blockers on this review.

### The one worth a ticket: a pull can silently discard local file edits

`syncProjectFiles` (`packages/cli/src/commands/sync.ts`) decides whether to download from the cloud side alone — it compares the inventory's `size`/`updatedAt` against the manifest record and never checks whether the local file was modified. So:

1. You edit a working file in a synced project (say `.files/data/notes.csv`).
2. Someone else changes that same file in Deepnote.
3. The next `sync --all-files` replaces your copy with theirs. No prompt, no warning, nothing in `-o json`.

Two things sharpen it:

- **A file edit on its own never uploads.** File sync follows the notebook direction (`outcome.action === 'pushed' || currentRecord.pendingFileUploads?.length`), so a file-only change sits on disk unsent until some notebook also changes — waiting to be overwritten.
- **The protection is one-directional.** Notebooks get override-or-skip in both directions, and this PR gives working files the same treatment on push. The download side has what it needs to do the same — the manifest record already stores `hash` — it just doesn't consult it.

This PR's wording change ("To accept the Deepnote versions, pull — this replaces your local copies") sets expectations correctly and is a real improvement. A download-side dirty check routed through the same `resolveConflict` would close the gap itself.

**Realistic severity: moderate.** It needs both sides to touch the same file, so it isn't an everyday event — but it's silent and unrecoverable without git.

### The rest — all low

| Issue | Plain version | Severity |
| --- | --- | --- |
| `updatedAt` has whole-second resolution | Change detection can't distinguish two versions of the same file written within one second at the same byte length. Both the pull unchanged-check and the divergence checks compare `updatedAt` + `size`, so such a pair looks identical | **Low** — I couldn't reproduce it in 8 attempts; a delete-then-upload round-trip costs roughly a second, and same-size is also required |
| Upload response may omit `size`/`updatedAt` | Both are optional in the create-file response. When they're absent, that file's baseline is unverifiable: it re-downloads on the next sync and isn't covered by the divergence checks until a pull refreshes it | **Low** — never observed in practice, bounded to the affected path, and self-healing |
| File/folder name shadowing | The file store will accept a file at `_deepnote_static/app` *and* a file at `_deepnote_static/app/index.html` at the same time. `publish --prune` clears the shadowing file; without `--prune` both persist and the CLI can't detect it | **Low, and really a question** — worth someone confirming what the static site serves in that state |

### Repo debt, not user-facing

- `packages/cli/src/integrations/parse-integrations.test.ts` — `getDefaultIntegrationsFilePath` asserts POSIX separators, so it always fails on Windows (`'\path\to\project\.deepnote.env.yaml'` vs `'/path/to/project/.deepnote.env.yaml'`). A Windows contributor can't get a green run, which trains people to ignore red.
- `packages/cli/src/commands/stats.test.ts` — "displays project name" fails intermittently on a clean checkout.

---

Method, for what it's worth: these came out of driving the real CLI against a mock API, with the mock's assumptions then validated against a live staging workspace. The first item reproduces deterministically; the severity ratings on the rest reflect how hard I found them to actually trigger, not just whether the code path exists.

packages/cloud/src/sync.test.ts
```diff
@@ -512,8 +512,10 @@ describe('downloadProjectFile', () => {
...
     ).rejects.toMatchObject({ statusCode: 413 })
   })
+})
 
-  it('defaults buffered working-directory transfers to 100 MiB', () => {
-    expect(MAX_BUFFERED_PROJECT_FILE_BYTES).toBe(100 * 1024 * 1024)
-  })
+// The CLI's own tests mock @deepnote/cloud, so this is the only assertion tying the exported
+// constant to the prefix the server actually serves static sites from.
+it('exports the static root the server serves from', () => {
```
tkislan: this test doesn't make any sense here

packages/cli/src/commands/publish.test.ts
```diff
@@ -6,6 +6,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 vi.mock('@deepnote/cloud', () => ({
   deleteProjectFile: vi.fn(),
   getProjectDetail: vi.fn(),
+  PROJECT_STATIC_ROOT: '_deepnote_static',
```
tkislan: you can use `vi.importActual` here .. you can check other tests as well

packages/cli/src/commands/publish.ts
```diff
@@ -4,16 +4,25 @@ import {
   deleteProjectFile,
   getProjectDetail,
   type ProjectStaticFilesUpdate,
+  PROJECT_STATIC_ROOT as STATIC_ROOT,
```
tkislan: I would just leave it `PROJECT_STATIC_ROOT` here .. makes it easier to search, and there's no real reason to actually rename here (I assume the goal was to avoid diffs, but that's negligible here)

skills/deepnote/references/apps.md
```diff
@@ -0,0 +1,168 @@
...
+Ownership note: the static root is a subtree of the same project file store that
+`deepnote sync --all-files` mirrors, so both commands write those paths. They share one baseline
+rather than splitting the namespace, and `publish` is the deploying writer — build output goes
+through `publish`, not through a synced workspace. Two consequences for anyone scripting a deploy:
+
+- Publish looks upwards from the published directory for a sync workspace and updates its mirror, so
+  the next sync sees the deploy as already in step. A CI deploy has no workspace to update and
+  should pass `--no-sync-root`.
+- If Deepnote holds changes below the target path that the workspace has not synced, publish exits 1
+  without touching the project. Sync first, or pass `--force`.
```
tkislan: this should match the wording from `cli-publish.md` file
```suggestion
- If a path publish is about to write or prune has moved on in Deepnote since that workspace last
  synced, it exits 1 without touching the project. Sync first, or pass `--force`. A sibling under
  the target that is not in this build (and not being pruned) does not stop the deploy.
```

packages/cloud/src/sync.test.ts
voyti: Right, test removed

packages/cli/src/commands/publish.test.ts
voyti: Done, in static-site-access.test.ts too.

packages/cli/src/commands/publish.ts
voyti: Done

skills/deepnote/references/apps.md
voyti: Applied as suggested

tkislan (approved):

